PR #8, opened 2026-01-08
Implement "roborev update" command and tui new version notification, incidental fixes 

### LINKED ISSUE #7: Add `roborev update` command
Would be a nice helper to upgrade when new releases are published.

A prompt that there's a new release on tui startup might also be nice.

wesm:
Fixed #7. Still some testing / QA needed


wesm:
cc @criccomini

PR #16, opened 2026-01-09
Add configurable job timeout setting

andyxhadji:
I've been using this to build a clinical data extraction workflow with a custom agent, and it's useful to have longer-running agent jobs! Thought I would contribute this piece back.

Add job_timeout_minutes to both global and per-repo config, with ResolveJobTimeout() function following the same priority pattern as ResolveAgent(). Default timeout remains 10 minutes (preserving existing behavior).

Priority order:
1. Per-repo config (.roborev.toml)
2. Global config (~/.roborev/config.toml)
3. Default (10 minutes)

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -177,6 +177,15 @@ func TestResolveJobTimeout(t *testing.T) {
 		}
 	})
 
+	t.Run("negative global config falls through to default", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


wesm:
I changed the default to 30 (I've come close to hitting 10 minutes even on small reviews!), this is great though, thanks!

PR #30, opened 2026-01-11
Add Nix flake for building and development

jzila:
## Summary
- Add `flake.nix` with default package that builds both `roborev` and `roborevd` binaries
- Include dev shell with Go toolchain (`go`, `gopls`, `gotools`)
- Configure `nativeCheckInputs` with git so tests pass during build

## Test plan
- [x] `nix build` succeeds and produces both binaries
- [x] Tests pass during build phase

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,4 +59,4 @@ jobs:
         run: nix build
 
       - name: Verify binary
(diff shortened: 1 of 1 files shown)


wesm:
thanks! just added CI to keep this healthy

wesm:
thank you!

PR #34, opened 2026-01-11
Fix git worktrees being treated as separate repositories

### LINKED ISSUE #32: Git worktrees are treated as separate repositories
## Problem

When using git worktrees, roborev incorrectly treats each worktree as a separate repository instead of recognizing it as part of the parent repo.

### Example
A worktree named `beads-sync` that's part of the `canopy` repo shows up as a separate "beads-sync" repository in the review list:

```
  ID Ref       Repo         Agent  Status
  4  0f5b545   beads-sync   codex  failed    <-- Should be "canopy"
  3  ca4fed3   canopy       codex  canceled
```

## Root Cause

The issue is in how repository paths are resolved and stored:

1. **`internal/git/git.go:111-122`** - `GetRepoRoot()` uses `git rev-parse --show-toplevel`, which returns the worktree's logical root directory, not the main repository root.

2. **`internal/storage/repos.go:31`** - Repository name is derived using `filepath.Base(absPath)`, which for a worktree returns the worktree directory name instead of the actual repo name.

3. **Database schema** - Uses `root_path` as the unique key, so worktrees (with different paths) create duplicate repo entries.

## Suggested Fix

Git provides `git rev-parse --git-common-dir` which returns the common git directory. For worktrees, this points to the main repo's `.git` directory. The fix would:

1. Add detection for worktrees using `--git-common-dir`
2. Resolve worktrees to their parent repository's root path
3. Use the parent repo path for storage lookups

Example implementation approach:
```go
// GetMainRepoRoot returns the main repository root, resolving through worktrees
func GetMainRepoRoot(path string) (string, error) {
    // Get the common git dir (works for both regular repos and worktrees)
    cmd := exec.Command("git", "rev-parse", "--git-common-dir")
    cmd.Dir = path
    out, err := cmd.Output()
    if err != nil {
        return "", err
    }
    gitDir := strings.TrimSpace(string(out))
    
    // For regular repos, --git-common-dir returns ".git" or absolute path to .git
    // For worktrees, it returns the main repo's .git directory
    if !filepath.IsAbs(gitDir) {
        gitDir = filepath.Join(path, gitDir)
    }
    
    // The repo root is the parent of the .git directory
    return filepath.Dir(gitDir), nil
}
```

## Default Behavior Consideration

Additionally, the default behavior should probably be to **only process commits from the main worktree**, not from linked worktrees. Worktrees are often used for temporary parallel work, experiments, or build artifacts where automatic reviews may not be desired. 

This could be:
- A config option like `ignore_worktrees = true` (default)
- Or require explicit `roborev init` in a worktree to enable reviews there

jzila:
## Summary
- Adds `GetMainRepoRoot()` function that uses `git rev-parse --git-common-dir` to resolve worktrees to their main repository
- Updates the daemon's enqueue handler to use this function for repo lookups
- Ensures commits from any worktree are correctly associated with the parent repository

Fixes #32

## Test plan
- [x] Added unit tests for `GetMainRepoRoot()` covering regular repos, worktrees, and error cases
- [x] All existing tests pass
- [ ] Manual testing: create a worktree, make a commit, verify it shows under the main repo name

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -158,8 +158,25 @@ func GetMainRepoRoot(path string) (string, error) {
 	// Only apply worktree resolution if git-dir differs from common-dir
 	// This ensures submodules (where both are the same) use GetRepoRoot
 	if gitDir != commonDir {
-		// This is a worktree - return parent of common dir (main repo root)
-		return filepath.Dir(commonDir), nil
+		// This is a worktree. For regular worktrees, commonDir ends with ".git"
+		// and the main repo is its parent. For submodule worktrees, commonDir
+		// is inside .git/modules/ and we need to read the core.worktree config.
+		if filepath.Base(commonDir) == ".git" {
+			// Regular worktree - parent of .git is the repo root
+			return filepath.Dir(commonDir), nil
+		}
+
+		// Submodule worktree - read core.worktree from config
+		cmd := exec.Command("git", "config", "--file", filepath.Join(commonDir, "config"), "core.worktree")
+		out, err := cmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("git config core.worktree for submodule worktree: %w", err)
+		}
+		worktree := strings.TrimSpace(string(out))
+		if !filepath.IsAbs(worktree) {
+			worktree = filepath.Join(commonDir, worktree)
+		}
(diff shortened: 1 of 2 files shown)


wesm:
I just rebased this to contain only this change. I'm reviewing it now

Issue #151, opened 2026-01-28
How do you have your main agent check reviews?


lucharo:
Hey @wesm, how do you have your main agents get or check the reviews for commits?

I like that the reviews are passive, but it'd be nice if there was a way to get a notification or hook to the main agent. I've found this frustrating with Claude Code for a while. It clearly has internal infrastructure to support notifications, given that when there is a background job or subagent running the main agent gets notified when it finishes along with an exit code. But there's no way for external tools to tap into that.

Currently I have instructions in my CLAUDE.md explaining what roborev is, how to use it, plus the skills that come with the CLI. But I still find Claude forgetting to check the reviews or forgetting to mark them as addressed.

I'm really curious to know what your setup and workflow is, and whether there might be an option to set up notifications in the future. Is polling the intended pattern?

wesm:
hi @lucharo — I always have at least one instance of the TUI open and visible and with the hide-addressed filter `h` active so I only see unaddressed reviews. Then I either use `/roborev:address $id` or press `y` to paste one or more unaddressed reviews into my Claude session prompt. I am running 4 - 5 Claude sessions at all times and so a lot of unaddressed reviews pile up. Are you using the TUI (you mentioned not being able to know when a review is finished so I guess not)? 

If you have different projects in progress on different desktop workspaces, it can be useful to have each TUI instance filtered down to a project and/or branch so you only see incoming reviews in your roborev "inbox" (queue).

I'm very much in "inbox zero" mode with my review queue (2755 reviews run already) and if I choose not to act on a review I mark it addressed so it drops out of the "inbox".

I don't know how to hook into Claude Code's built-in notification infrastructure but if someone knows I'd be interested to do more there.

wesm:
I'll close this but feel free to propose specs for specific features to build

PR #179, opened 2026-01-30
feat: add Claude Code PostToolUse hook for review integration

lucharo:
Hi @wesm, as we discussed briefly in #151, there's currently no way for the main agent to know that a `roborev` review is happening in the background.

I had been using the `roborev` TUI, but I wanted to remove myself from the loop and make the agent aware - something close to a notification system. The Claude Code team hasn't released SDK support for this or a way to tap into their notification system (which seems to work for background tasks), so this is my attempt.

The hook at `internal/hooks/claude/roborev-post-commit/hook.sh` triggers after git commits, queues a review, and outputs a reminder for the agent to address findings.

A few tricks worth noting:

1. Hooks with exit code 0 don't pass context to the agent (useful for linting, not for our use case here). Exit code 2 passes the hook output to the main context. A bit of a magic number, but it works. (see https://code.claude.com/docs/en/hooks#simple:-exit-code)

2. Hook matchers can't use argument patterns like the permissions system can - `Bash(git commit*)` doesn't work, only `Bash`. So the hook fires on every Bash call and exits early (with 0) when it's not a git command. I raised [anthropics/claude-code#21540](https://github.com/anthropics/claude-code/issues/21540) about this limitation but I don't think it'll be addressed.

3. The pattern matches "git commit" anywhere in the command to catch chained commands like `git add && git commit`. Tradeoff: occasional false positives from heredocs or string arguments containing "git commit".

4. The hook output is kept lean to avoid polluting the agent's context.

Go code changes are still needed to embed and install it to `~/.claude/hooks/`.

## Question

How should installation work?

1. New command: `roborev hooks install`
2. Extend existing: `roborev skills install --hooks`
3. Something else?

### DIFF
--- a/internal/hooks/claude/roborev-post-commit/hook.sh
+++ b/internal/hooks/claude/roborev-post-commit/hook.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# Claude Code PostToolUse hook for roborev integration
+# Triggers after git commit commands to remind the agent to check reviews
+
+set -euo pipefail
+INPUT=$(cat)
+
+# Only trigger on git commit (match command field specifically, not full JSON)
+echo "$INPUT" | jq -r '.tool_input.command // ""' | grep -q "git commit" || exit 0
+
+# Check prerequisites
+command -v roborev &>/dev/null || exit 0
+git rev-parse --git-dir &>/dev/null || exit 0
+roborev status &>/dev/null || exit 0
+
+COMMIT_SHA=$(git rev-parse HEAD)
+COMMIT_SHORT=$(git rev-parse --short HEAD)
+
+# stderr + exit 2 = Claude sees this message
+echo "**ROBOREV** — Review queued for ${COMMIT_SHORT}. Run \`sleep 25 && roborev show ${COMMIT_SHA}\`, note job ID, then \`roborev address <id>\`." >&2
+exit 2


wesm:
Interesting proposal. I'll have to set this up and see what the ergonomics look like. Personally I don't want finished reviews randomly hijacking my interactive sessions -- if I want to use an interactive session to address, I prefer to use the TUI to paste multiple reviews (with `y`) into the prompt to have discretion about which reviews are addressed at any given time. Of course now with `roborev fix` and `roborev refine` I'm using interactive sessions less and less to address reviews. 

So if users want what you are doing, we can give them a way to have it, but I'm unlikely to adopt this workflow myself.

lucharo:
Thanks for the very quick reply!  several questions/thoughts

---

I guess it really depends on developer and especially git workflow. I tend to iterate on an idea and only commit when I am confident something has been achieved and isn't wrong so a review after each commit is rarely noise and in general keeps the main agent from diverging. I guess for users that commit every often that might not work as well. 

My workflow was already asking the main agent to spawn a sub-agent and then address the review. So `roborev` in the way I was using was a much leaner autonomous version of that 

--- 

for your workflow, I still don't understand the `y` workflow, do you select a few, like in vim (e.g. `5y`), then yank those reviews and have the agents address? 

--- 

wow! `roborev {refine,fix}` are super cool! Have you thought of adding a github action that runs `roborev` refine when an issue is created? It's amazing, might change how I was currently envisioning using `roborev` — will keep playing with it!

wesm:
> I tend to iterate on an idea and only commit when I am confident something has been achieved and isn't wrong so a review after each commit is rarely noise and in general keeps the main agent from diverging. I guess for users that commit every often that might not work as well.

I feel very strongly that this is the wrong way to work now and rooted in the old human way of working on something until it's "ready" and then committing. After working on this project for a month, I've concluded that you want other agents looking at the code that is being generated *right away*, not after you've iterated and it's "done" (because it's not done -- it has lots of bugs!)

> for your workflow, I still don't understand the y workflow, do you select a few, like in vim (e.g. 5y), then yank those reviews and have the agents address?

I dunno, the copy pasting seems intuitive to me? here is a screencast of me doing it

https://github.com/user-attachments/assets/09e9091e-2efe-405d-9f5d-247c67fa4501

> Have you thought of adding a github action that runs roborev refine when an issue is created? It's amazing, might change how I was currently envisioning using roborev — will keep playing with it!

Yeah starting to think about it! all in good time

lucharo:
Oh I see! `y` yanks the whole review I see, I do get what you're saying of the old vs new way of working, in that same way isn't it quite noisy to have agents review every single commit even though a piece of work might not be complete? Like, with `roborev` do you only address a review every x commits? wouldn't it make sense to address each commit no matter often they are made? 

Just to show the workflow in action: 

```txt
⏺ Bash(git add .github/workflows/automated-release.yml CHANGELOG.md CLAUDE.md tests/test_submission.py
      vertex_cli/submission.py && git commit -m "feat: output console…)
  ⎿  [main 67e2d20] feat: output console URL after submission, fix release workflow skip
      5 files changed, 79 insertions(+), 6 deletions(-)
     ruff.....................................................................Passed
     ty.......................................................................Passed
  ⎿  PostToolUse:Bash hook returned blocking error

⏺ Bash(sleep 25 && roborev show 67e2d20c4965e796da24ac5e3d46561945c258c8)
  ⎿  Review for 67e2d20 (job 49, by claude-code)
     ------------------------------------------------------------
     ## Summary
     … +58 lines (ctrl+o to expand)

⏺ Good catches. Let me fix all three:
```

---

Also, I guess, just to note, I still commit very often, and 99% of commits are done by claude code, I still have it commit in "logical units of work" though that may as well be outdated.

wesm:
My codex reviews typically take anywhere from 15s to 5 minutes, so it doesn’t make sense for the agent to stop and wait for review when it could be doing other things while waiting on the review. The asynchronous model is simply more productive in my experience, it sounds like you want a synchronous workflow but I’m quite confident this is slow.

wesm:
> isn't it quite noisy to have agents review every single commit even though a piece of work might not be complete?

In my experience so far (> 3000 reviews processed), this isn't really an issue. If the reviews raise an issue that has already been addressed, the agents ignore it. They always look at the current state of the codebase and so changes that have occurred since the review was run does not appear to affect their performance either

wesm:
I think the review hooks system in #181 is the way to go and keep the focus on asynchronous reviews. If someone wants their interactive Claude sessions to block and wait for a review, we should document the approach for that in https://github.com/roborev-dev/roborev-docs

lucharo:
To be honest, I tried to implement an async notification hook where Claude would launch the `roborev` command to check the latest review in the background but couldn't get it to work — again this improve if Claude code or other agent CLIs better expose their notification system. 

--- 

I was using `roborev` in the way you suggest and I still find the the stream of reviews to be a bit too much, especially if the agent is not aware of them. I understand that reviews are aware of n-3 reviews/code changes though I wasn't aware initially. I'd still find myself feeding all the reviews after several commits to the main agent. 

--- 

Your workflow sounds fascinating though. I still can't imagine what using agents fully non interactively looks like, I can't trust them to not diverge very quickly and waste a bunch of time after. Also you saying your average  review can be up to 5 mins made me reconsider whether I'm not being ambitious enough with my agents tasks (averaging 1-3 mins reviews after serious project use) 


--- 

**either way**, I don't want to clutter a PR with a discussion! please feel free to close the PR if it doesn't go with the intended use/design philosophy of the project. I'll keep iterating and finding ways the tool fits my workflow, thanks @wesm!

wesm:
All good, I'm very curious about what works well for different people! I'll investigate how to create some kind of a "review inbox" for Claude since I don't think that blocking the interactive session makes sense as a design. Here are some statistics on runtimes for the reviews in my environment:

```
⏺ ┌──────────────┬─────────┬───────────┬────────┬─────┬─────┬─────────┐                                                                                                                                  
  │    Agent     │ Reviews │ Completed │ Failed │ Avg │ Min │   Max   │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ codex        │ 2,696   │ 2,661     │ 9      │ 93s │ <1s │ 19m 55s │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ gemini       │ 500     │ 496       │ 1      │ 69s │ <1s │ 9m 53s  │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ claude       │ 479     │ 477       │ 1      │ 14s │ <1s │ 4m 55s  │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ opencode     │ 9       │ 2         │ 7      │ 12s │ <1s │ 1m 46s  │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ ollama       │ 8       │ 0         │ 8      │ <1s │ <1s │ 1s      │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ droid        │ 4       │ 1         │ 3      │ 15s │ 1s  │ 58s     │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ copilot      │ 2       │ 2         │ 0      │ 63s │ 61s │ 64s     │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ gpt-5.1-nano │ 1       │ 1         │ 0      │ 9s  │ 9s  │ 9s      │                                                                                                                                  
  ├──────────────┼─────────┼───────────┼────────┼─────┼─────┼─────────┤                                                                                                                                  
  │ cursor       │ 1       │ 1         │ 0      │ 24s │ 24s │ 24s     │                                                                                                                                  
  └──────────────┴─────────┴───────────┴────────┴─────┴─────┴─────────┘ 
```

I think some of the longer reviews were scenarios when I was connected to my mobile hotspot or on an airplane, but this should give an idea why the synchronous review concept is problematic

wesm:
> I'd still find myself feeding all the reviews after several commits to the main agent.

FWIW, when I'm not using `roborev fix`, I'm generally feeding batches of 3-5 reviews to the interactive session through the prompt with the `y` key like I showed you in the video. For me, this is yielding very good results in terms of finding and fixing issues, and avoiding agent idleness (and as a result I'm hitting my weekly rate limit on Claude at least twice as fast as I was before).

wesm:
I have taken a closer look at this and I'm going to close this PR since synchronous review is at odds with the design of the project, which is for reviews to be asynchronous and the new `/roborev:fix` skill allows you to "fix all unaddressed reviews" whenever you want. I think that keeping an eye on the TUI and explicitly flipping between code generation turns and addresses reviews using the skills is the intended way to use this project, with the goal of the agent always staying busy doing something (building new things, or cleaning up its own messes).

Now with the new hooks system (https://www.roborev.io/guides/hooks/), you have the ability to customize what happens on review completion. 

I asked Claude Code to weigh in on this conundrum and here is what it said:

```
Honestly, I don't think there's a good way to do it with current
Claude Code architecture. The fundamental problem is that Claude Code
has no "push notification" mechanism — external events can't interrupt
an active session. The only injection points are hooks that fire on
tool calls or user prompts.

The options I can think of all have significant drawbacks:                                                                                                                                             
File-based inbox with hook polling: on_review_completed appends to a
notifications file, and a Claude Code PostToolUse hook checks it on
every tool call. But this fires on every tool call — reading files,
grep, edits — which is extremely noisy. You could throttle it (only
check if file was modified recently), but then you're building a
fragile polling system inside shell hooks.

Post-commit check for completed reviews: Instead of sleep 25, check if
there are already-completed unaddressed reviews for the repo. This
avoids the sleep but only fires at commit time — the agent wouldn't
learn about reviews that complete while it's doing non-commit work.

User-prompt-submit hook: Check for pending reviews when the user sends
message. Less noisy than every tool call, but still only fires on user
interaction, not on review completion.

All of these are working around the fact that Claude Code hooks are
reactive (triggered by tool calls) not event-driven (triggered by
external events). You'd be building a poll-on-every-action system
disguised as notifications.

The async workflow you already have — TUI for browsing, y to yank
reviews into the prompt, roborev fix for automated addressing — is
genuinely better suited to how Claude Code works. The agent stays
productive instead of waiting, and you control when reviews enter the
conversation.

If Claude Code ever adds a proper notification/inbox API that external
processes can push messages into, that would be the right integration
point. But that's a Claude Code feature request, not something roborev
can solve on its own.
```

I appreciate the dialogue and perhaps systems like roborev will help nudge the Claude Code team to develop some new hook points that will enable this tools to work better together in the future.

wesm:
Food for thought: https://github.com/mrocklin/claudechic/pull/36

PR #187, opened 2026-02-01
Add {findings} template variable to hook commands

jzila:
## Summary
- Adds a `{findings}` template variable to hook command interpolation, containing the full review output from the agent
- The variable is shell-escaped like all other template variables to prevent injection
- Populated on `review.completed` events

## Test plan
- [x] Existing interpolation tests updated and passing
- [x] Shell injection tests extended to cover `{findings}`
- [x] Full test suite passes (`go test ./...`)


wesm:
thanks!

PR #195, opened 2026-02-03
Feature fix list fixes available

hughdbrown:
Add an option that so that, before applying a fix, there is a list of the available fixes.


hughdbrown:
I don't get this:
> 1 workflow awaiting approval
> This workflow requires approval from a maintainer.

Is there something here I have to fix? Why is it not green?

wesm:
> Is there something here I have to fix? Why is it not green?

I have to approve github actions runs from first-time contributors (I am going to tweak the settings, but I believe your PRs will run automatically afterward)

wesm:
image

PR #197, opened 2026-02-04
Fix wrong agent and duplicate reviews from post-commit hook

### LINKED ISSUE #196: Bug: roborev calls review / performs second review with the wrong agent
I use Claude, not Codex. I keep getting Codex called to perform fixes.

Apply fixes with Claude:
```
10054* 2026-02-03 09:47  roborev fix --agent claude 97
10058* 2026-02-03 09:55  roborev fix --agent claude 99
```

Get successful fixes with Claude *and* failed fixes with Codex.
```
roborev queue (v0.23.1)
Daemon: v0.23.1 | Workers: 0/4 | Done: 96 | Addressed: 2 | Unaddressed: 94

  ID  Ref              Branch                    Repo                       Agent        Status   P/F Queued       Elapsed  Addressed
  -----------------------------------------------------------------------------------------------------------------------------------
  100 78151f9          feature-fix-list-fixes... roborev                    codex        failed   -   Feb 03 09:57 8s
  99  1c1508c          feature-fix-list-fixes... roborev                    claude       done     P   Feb 03 09:53 16s      true
  98  1c1508c          feature-fix-list-fixes... roborev                    codex        failed   -   Feb 03 09:50 7s
> 97  5f959cc          feature-fix-list-fixes... roborev                    claude       done     F   Feb 03 09:39 18s      true
```

wesm:
## Summary

Fixes two problems when a fix agent creates a commit and the post-commit hook fires:

1. **Wrong agent**: The hook sends no `--agent` flag, so the server defaulted to `codex` even when codex wasn't installed. The enqueue handler now resolves the configured agent to the first **installed** agent via the fallback chain (codex -> claude-code -> gemini -> ...), so the job reflects what will actually run.

2. **Duplicate reviews**: Both the post-commit hook and `enqueueIfNeeded` (called after a fix completes) could create separate reviews for the same commit. Replace the fixed 2s sleep with a **10 x 1s polling loop** that returns as soon as the hook's job is found, plus a final check after the loop.

Also returns **503** when no review agent is installed at all, failing fast instead of queuing a doomed job.

## Files changed

- `internal/daemon/server.go` — resolve agent to first available at enqueue time, 503 on none
- `cmd/roborev/fix.go` — polling loop in `enqueueIfNeeded`
- `internal/daemon/server_test.go` — 5 agent availability subtests with cross-platform PATH isolation

## Test plan

- [x] `go test ./internal/daemon/ -run TestHandleEnqueueAgentAvailability -v` — all 5 subtests pass
- [x] `go test ./cmd/roborev/ -run TestEnqueueIfNeeded -v` — all 3 subtests pass
- [x] `go test ./...` — full suite passes

Fixes #196

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -1013,6 +1013,9 @@ func enqueueIfNeeded(serverAddr, repoPath, sha string) error {
 		}
 		time.Sleep(1 * time.Second)
 	}
+	if hasJobForSHA(serverAddr, sha) {
+		return nil
+	}
 
 	branchName := git.GetCurrentBranch(repoPath)
 


wesm:
cc @hughdbrown

wesm:
I don't think this is a complete fix, I'm still investigating

hughdbrown:
@wesm Dogfooded problems reported by roborev:

```
### Medium Severity

**1. Shell injection vulnerability in wrapper scripts**
- File: `internal/daemon/server_test.go:2603-2609`
- Both the Windows and Unix wrapper scripts directly interpolate `gitPath` into the script content without escaping. If the git
path contains special characters (single quotes on Unix, double quotes or percent signs on Windows), the wrapper could malfunction
or potentially execute unintended commands.
- **Unix example**: A path like `/usr/local/bin/git's-version` would break the shell script due to the unescaped single quote.
- **Windows example**: A path containing `%` or `"` could cause issues.
- **Suggested fix**: Escape the path appropriately for each platform:
  ```go
  if runtime.GOOS == "windows" {
      // Escape double quotes for Windows batch files
      escapedPath := strings.ReplaceAll(gitPath, `"`, `""`)
      wrapper := fmt.Sprintf("@\"%s\" %%*\r\n", escapedPath)
      // ...
  } else {
      // Escape single quotes for shell scripts
      escapedPath := strings.ReplaceAll(gitPath, "'", "'\"'\"'")
      wrapper := fmt.Sprintf("#!/bin/sh\nexec '%s' \"$@\"\n", escapedPath)
      // ...
  }
  ```
- **Note**: In practice, git is typically installed in a standard path without special characters, so this is unlikely to cause
actual CI failures, but it's a defensive coding concern.
```

hughdbrown:
@wesm And also:

```
### Medium Severity

**1. Semantic change in availability checking may cause unexpected 503 errors**
- File: `internal/daemon/server.go:454-458`
- The old code called `agent.GetAvailable(agentName)` and used the result to potentially update `agentName` to an available
fallback. The new code still calls `GetAvailable` but now returns 503 on error.
- Looking at the test case "unavailable codex preserved when fallback exists" - it expects `codex` (unavailable) to be preserved
and return 201, with only `claude` in PATH. This implies `GetAvailable("codex")` must succeed when `claude` is available (i.e., it
internally falls back).
- **Potential issue**: If `GetAvailable` returns an error when the *specific* requested agent is unavailable (rather than when
*all* agents are unavailable), then requesting an unavailable agent would now return 503 even when fallbacks exist. The test
suggests this works correctly, but the logic depends on `GetAvailable`'s internal behavior not being obvious from this diff.
- **Suggested fix**: Add a comment clarifying that `GetAvailable` returns success if *any* agent (including fallbacks) is
available, not just the requested one. Or consider using a separate function like `agent.AnyAvailable()` for the fail-fast check to
make the intent clearer.

### Low Severity

**2. Test case "default codex preserved when fallback exists" may have misleading name**
- File: `internal/daemon/server_test.go:2627-2632`
- When `requestAgent` is empty (`""`), the test expects `expectedAgent: "codex"`. This suggests `config.ResolveAgentForWorkflow`
returns `"codex"` as the default. The test name says "default codex preserved" but it's really testing that whatever the config
resolves to is preserved.
```

hughdbrown:
@wesm And finally:

```
### High Severity

**1. Silent failure when no agents are available**
- File: `internal/daemon/server.go:454-457`
- When `agent.GetAvailable()` returns an error (meaning no agents are available), the code silently continues with the original
`agentName`, which was just validated as unavailable. This will create a job with an agent that cannot execute.
- **Suggested fix**: Return an error to the client when no agent is available:
  go
  if resolved, err := agent.GetAvailable(agentName); err == nil {
      agentName = resolved.Name()
  } else {
      http.Error(w, fmt.Sprintf("no available agent: %v", err), http.StatusServiceUnavailable)
      return
  }
  

### Medium Severity

**2. Missing test case for "no agents available" scenario**
- File: `internal/daemon/server_test.go:2583-2683`
- The tests cover various fallback scenarios but don't test what happens when *no* agents are available (empty `mockBinaries` with
a non-test agent). Given the bug above, this would currently create a job that will fail at execution time.
- **Suggested fix**: Add a test case:
```

wesm:
@hughdbrown most of those review findings are addressed in the commits I just pushed. regarding the shell injection concern:

```
  Comment 1 — "Shell injection in wrapper scripts" (Medium)                                         
  This is the only finding that applies to the current code. gitPath from exec.LookPath("git") is   
  interpolated directly into wrapper scripts. Hugh notes himself it's "unlikely to cause actual CI  
  failures" since git lives in standard paths (/usr/bin/git, /opt/homebrew/bin/git, C:\Program      
  Files\Git\cmd\git.exe). This is test-only code constructing a wrapper to a path the OS just gave  
  us — I'd call this a non-issue in practice. Not worth the escaping complexity.   
```

wesm:
Could you let me know if this branch resolves the issue that you were seeing before I merge it?

hughdbrown:
> Could you let me know if this branch resolves the issue that you were seeing before I merge it?

Testing:
- git pull upstream no-codex-multiple-reviews
- vim ~/.roborev/config.toml (set default-agent to empty string)
- find a change to make, make a branch for the changes, run review with `--agent claude`, confirm that roborev does not try to run codex/some agent I don't have

This will take a minute.

hughdbrown:
```
commit 2a065c47315a9458cf3b398236bd1befbda4f66e
Refs: [chore-ignore-bin-dir], v0.23.1-8-g2a065c4
Author:     Hugh Brown <hughdbrown@yahoo.com>
AuthorDate: Tue Feb 3 20:07:48 2026 -0700
Commit:     Hugh Brown <hughdbrown@yahoo.com>
CommitDate: Tue Feb 3 20:07:48 2026 -0700

    These are not the droids you are looking for
---
 .gitignore | 1 +
 1 file changed, 1 insertion(+)

diff --git a/.gitignore b/.gitignore
index 8013f91..ff6e718 100644
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.exe
 /dist/
 bin/
+floogle-non-existent-dir/
```

```
$ roborev review --agent claude 2a065c4
...
Enqueued job 123 for 2a065c4 (agent: claude)

$ roborev tui
```

```
**Low** - `.gitignore` - Unclear purpose of ignored directory

The directory name `floogle-non-existent-dir/` appears to be a placeholder or test entry with no clear purpose. The commit message
("These are not the droids you are looking for") is whimsical but doesn't explain why this entry is needed.

```

And then:
```
$ roborev fix 123
...
stream error: Failed to refresh token: 401 Unauthorized; retrying 1/5 in 205ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 2/5 in 406ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 3/5 in 796ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 4/5 in 1.465s…
stream error: Failed to refresh token: 401 Unauthorized; retrying 5/5 in 3.2s…
ERROR: Failed to refresh token: 401 Unauthorized
```

**So, no, this does not fix the original manifestation.**

hughdbrown:
Argh:

```
~/workspace/hughdbrown/roborev chore-ignore-bin-dir
```

Tested on the wrong branch.

hughdbrown:
Try again:
```
~/w/hu/roborev no-codex-multiple-reviews cherry ~1 ❯ vim .gitignore                                                       15s 20:17:22
~/w/hu/roborev no-codex-multiple-reviews cherry ~1 ❯ git add .gitignore                                                   13s 20:17:51
~/w/hu/roborev no-codex-multiple-reviews cherry +1 ❯ git commit -m "These are not the droids we are looking for"           ...
[no-codex-multiple-reviews **170fc4f**] These are not the droids we are looking for
...
~/w/hughdbrown/roborev no-codex-multiple-reviews ⇡1 ❯ roborev review --agent claude 170fc4f                                   20:18:25
zsh: correct 'claude' to '.claude' [nyae]? n
Enqueued job 124 for 170fc4f (agent: claude)
~/w/hughdbrown/roborev no-codex-multiple-reviews ⇡1 ❯ roborev tui                                                             20:18:48
~/w/hughdbrown/roborev no-codex-multiple-reviews ⇡1 ❯ roborev fix 124
...
stream error: Failed to refresh token: 401 Unauthorized; retrying 1/5 in 189ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 2/5 in 374ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 3/5 in 827ms…
stream error: Failed to refresh token: 401 Unauthorized; retrying 4/5 in 1.493s…
stream error: Failed to refresh token: 401 Unauthorized; retrying 5/5 in 2.934s…
ERROR: Failed to refresh token: 401 Unauthorized
...
```

Still no joy.

wesm:
That looks like a different problem, which is that you have `codex` installed, no tokens available, and you do not have `default_agent` set in config.toml, is that right? The best fix for your situation is to make sure you have `default_agent = "claude"` in ~/.roborev/config.toml

wesm:
```
  He has codex installed (so exec.LookPath("codex") succeeds and             
  IsAvailable("codex") returns true), but it's not authenticated — the 401 errors are from codex    
  trying to refresh an OAuth token.                                                                 
                                                                                                    
  So GetAvailable("codex") resolves to codex (it's "available" in the binary-exists sense), the job 
  gets stored with agent codex, the worker runs codex, and codex fails at runtime because it can't  
  authenticate.                                                                                     
                                                                                                    
  Our fix doesn't help here — IsAvailable only checks if the binary exists in PATH, not whether it's
   actually functional. His roborev fix 124 ran without --agent claude, so it defaulted to codex.   
                                                                                                    
  The post-commit hook path has the same problem: hook fires with no agent → resolves to codex      
  (installed) → worker runs codex → 401.                                                            
                                                                                                    
  This is a separate issue from what we fixed. Our change handles the case where the agent binary is
   missing. His case is where the binary exists but isn't configured. The workaround wesm suggested 
  (default_agent = "claude" in config) is the right answer for him — it bypasses the fallback chain 
  entirely.               
```

hughdbrown:
Huh:

```
~ ❯ which codex
/opt/homebrew/bin/codex
```

okay, I don't use it so I am unclear where it came from.

> The best fix for your situation is to make sure you have default_agent = "claude" in ~/.roborev/config.toml

Yeah, this is what you told me to do earlier today, and which I undid tonight to tee up the problem again for testing.

wesm:
I *think* this PR resolves the duplicate review problem, but if it recurs just open a new issue and we can work on it

hughdbrown:
> He has codex installed ...
> The workaround wesm suggested ...

I am getting schooled by some coding assistant? [Oh the humanity](http://www.nocaptionneeded.com/wp-content/uploads/2012/05/Hindenburg.shere_.jpg)

PR #210, opened 2026-02-05
Add jump-to-top shortcut to TUI

bstancil:
There is this thing that happens when you're looking at the the list of commits in the TUI: 

1. You hide all of the approved commits.
2. Now, some unapproved one is selected. 
3. You unhide the approved commits. 
4. The unapproved one is still selected. If it's a bunch of commits down the list, you have to arrow up a whole bunch to get to the top.

Now, you can just press g. 

And, most importantly:

Screenshot 2026-02-05 at 3 05 58 PM

(Plus, some linting).


bstancil:
dope, thank you!

PR #214, opened 2026-02-06
Add hide_addressed_by_default config, Gemini default model, agent/test hardening

anshbansal:
## Summary

- **TUI: read `hide_addressed_by_default` from config** — the TUI now reads `hide_addressed_by_default` from `~/.roborev/config.toml` on launch so the `h` filter can be enabled by default. Pressing `h` toggles the filter for the current session only; the config file is never written back.
- **Default Gemini agent to `gemini-3-pro-preview`** — previously no model was set, so the Gemini CLI fell back to flash.
- **`WithModel("")` preserves defaults across all agents** — previously calling `WithModel("")` on Gemini (and others) would clear the built-in default model. Now all six agents treat an empty string as a no-op.
- **Test isolation** — added `TestMain` in `cmd/roborev` to set `ROBOREV_DATA_DIR` to a temp dir, preventing tests from reading the developer's real config. Fixed temp dir leak where `defer` + `os.Exit` skipped cleanup.
- **CursorAgent test coverage** — added CursorAgent to `getAgentModel`, `TestAgentWithModelPersistence`, and `TestWithModelEmptyPreservesDefault` so all agent types are verified.
- **Config mutation regression tests** — added tests for both malformed and valid `config.toml` files asserting they are not mutated when toggling `h`.
- **Test helpers** — added `TempDir`, `Chdir`, and `assertArgsSequence` helpers; added explicit `MkdirAll` error checks in `repo_test.go`.

## Test plan

- [x] `go test ./...` passes
- [x] Set `hide_addressed_by_default = true` in config.toml, open TUI, confirm addressed items are hidden on launch
- [x] Press `h` to toggle off, confirm config file is unchanged
- [x] Remove the setting (or set to false), reopen TUI, confirm addressed items are visible on launch

### DIFF
--- a/cmd/roborev/testmain_test.go
+++ b/cmd/roborev/testmain_test.go
@@ -14,8 +14,8 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
-	defer os.RemoveAll(tmpDir)
-
 	os.Setenv("ROBOREV_DATA_DIR", tmpDir)
-	os.Exit(m.Run())
+	code := m.Run()
+	os.RemoveAll(tmpDir)
+	os.Exit(code)
 }
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -205,6 +205,8 @@ func getAgentModel(a Agent) string {
 		return v.Model
 	case *OpenCodeAgent:
 		return v.Model
+	case *CursorAgent:
+		return v.Model
 	default:
 		return ""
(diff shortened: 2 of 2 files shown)


wesm:
Sorry I'm kind of taking over this branch a little bit but I thank you for your contribution! I will merge this soon.

I run multiple instances of the TUI at any given time and so persisting the hide-addressed state does not make sense, but changing the default from false to true does.

PR #225, opened 2026-02-10
QoL: command display, prompt navigation, auto-filter, server-side filtering & metrics

wesm:
## Summary
- Display agent command line in prompt view, truncated to terminal width using `runewidth.Truncate`
- Compute command line from job params at render time instead of storing in DB
- Add left/right arrow navigation between prompts in prompt view
- Fix prompt view rendering bug (raw ANSI escape sequences conflicting with bubbletea alt screen)
- Show job ID in prompt view title
- Add `auto_filter_repo` config option to auto-filter TUI to current repo on startup
- Server-side `addressed` filtering via `hide_addressed` query param
- Server-side aggregate counts (done/addressed/unaddressed) via `CountJobStats` for accurate metrics across paginated views
- Fix queue underfill when hide-addressed filters reduce visible results below page size
- Add pagination from review/prompt views when navigating past loaded jobs
- Fix `auto_filter_repo` not showing repo filter in title bar and not being clearable via escape
- Skip failed jobs when navigating between prompts
- Guard against stale prompt responses with jobID matching
- Clear `loadingMore` on stale response discard
- Gate `paginateNav` on correct view and clear on exit/error

## Test plan
- [x] `go test ./...` passes
- [x] Prompt view renders correctly with left/right navigation
- [x] Pagination triggers from review/prompt views at end of loaded data
- [x] Auto-filter shows repo in title bar and clears with escape
- [x] Metrics (done/addressed/unaddressed) reflect all jobs, not just loaded page
- [x] Multi-repo filter falls back to client-side counting

### DIFF
--- a/cmd/roborev/tui.go
+++ b/cmd/roborev/tui.go
@@ -100,6 +100,10 @@ const (
 	tuiViewTail
 )
 
+// queuePrefetchBuffer is the number of extra rows to fetch beyond what's visible,
+// providing a buffer for smooth scrolling without needing immediate pagination.
+const queuePrefetchBuffer = 10
+
 // repoFilterItem represents a repo (or group of repos with same display name) in the filter modal
 type repoFilterItem struct {
 	name      string   // Display name. Empty string means "All repos"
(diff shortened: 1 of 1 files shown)


wesm:
```
  1. Secrets in command_line — No longer stored in DB at all; computed on the fly from agent       
  params. The agent implementations don't include API keys (those come from env vars).
  2. Pagination stuck with "(none)" — Already handled: "(none)" sets needsAllJobs=true so limit=0  
  is used and pagination guards prevent loadingMore from ever being set.                 
  3. Terminal escape injection — The command line is computed from controlled code paths (agent
  name + flags), not user input.
  4. UTF-8 byte slicing — Agent command lines are ASCII (binary names, flags, model names). The
  only non-ASCII is the ellipsis we append.
  5. Multi-repo limit=0 — Rare edge case, bounded by local SQLite DB size.
  6. Tests outdated — Already updated in prior commits.

  Nothing to do here.
```

PR #235, opened 2026-02-11
Fix range review failing when start is root commit

wesm:
## Summary

- When reviewing a range like `roborev review <root> <end>`, the CLI constructs `<root>^..<end>` to make it inclusive. This fails when `<root>` is the repo's first commit since it has no parent.
- The server now detects this case and substitutes the empty tree SHA as the range start, producing a correct diff that includes the root commit's changes.
- The fallback is gated on `^{commit}` resolution so non-commit objects (blobs, trees) are still properly rejected.

## Test plan

- [x] `TestHandleEnqueueRangeFromRootCommit` — root commit range enqueues successfully with empty tree SHA
- [x] `TestHandleEnqueueRangeNonCommitObjectRejects` — blob SHA with `^` suffix returns 400
- [x] Full test suite passes

### DIFF
--- a/internal/daemon/server.go
+++ b/internal/daemon/server.go
@@ -574,7 +574,7 @@ func (s *Server) handleEnqueue(w http.ResponseWriter, r *http.Request) {
 			// the range includes the root commit's changes.
 			if strings.HasSuffix(parts[0], "^") {
 				base := strings.TrimSuffix(parts[0], "^")
-				if _, resolveErr := git.ResolveSHA(gitCwd, base); resolveErr == nil {
+				if _, resolveErr := git.ResolveSHA(gitCwd, base+"^{commit}"); resolveErr == nil {
 					startSHA = git.EmptyTreeSHA
 					err = nil
 				}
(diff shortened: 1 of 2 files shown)


wesm:
```
No, none of these are real issues.                 

  Medium - HEAD^^..HEAD coerced to EmptyTreeSHA: The only way HEAD^^ reaches the server is if the  
  user ran roborev review HEAD^ <end>. The CLI adds ^ to make the range inclusive. If HEAD^ is the
  root commit, then HEAD^^ correctly doesn't resolve, and substituting EmptyTreeSHA is the right   
  behavior — it includes the root's changes. The fallback is working exactly as intended.        
                                                                                                
  Low - argument injection in ResolveSHA: The ref comes from the CLI which got it from user     
  positional args. The user already has shell access. Refs starting with - would also just fail to
  resolve.                                                                                       
                                                                                            
  Low - panic from parts[1]: SplitN(gitRef, "..", 2) is only reached after the isRange check on
  line 566, which already verified gitRef contains "..". Two parts are guaranteed.

  Low - missing test for ^^: As above, the ^^ case represents correct behavior (root commit's
  inclusive range), not an error to reject.
```

PR #242, opened 2026-02-11
TUI markdown rendering, escape sanitization, and test infrastructure

wesm:
## Summary

- Add glamour-based markdown rendering to TUI review and prompt views with caching, scroll fixes, and configurable tab width
- Sanitize non-SGR terminal escape sequences (OSC, DCS, private-mode CSI) in rendered output to prevent terminal injection from untrusted content
- Implement CommonMark-compliant fence detection for code block pre-truncation
- Move slow CLI tests behind `//go:build integration` tag (unit tests: 35s -> 12s)
- Separate Postgres tests into `//go:build postgres` tag to avoid CI failures in jobs without Postgres
- Restructure Makefile test targets: `test` (unit), `test-integration` (unit+slow), `test-postgres` (postgres), `test-all` (everything)
- Update CI to pass correct build tags per job

## Test plan

- [x] Unit tests pass (`go test ./...`)
- [x] Integration tests pass (`go test -tags integration ./...`)
- [x] Postgres tests pass (`make test-postgres`)
- [x] Escape sanitization covers SGR preservation, OSC/DCS/CSI stripping, private-mode CSI, and incomplete sequences
- [x] Fence detection handles CommonMark edge cases (indented fences, tilde fences, invalid closers)
- [x] Fallback render path applies sanitization

### DIFF
--- a/cmd/roborev/tui_helpers.go
+++ b/cmd/roborev/tui_helpers.go
@@ -288,9 +288,10 @@ var csiRe = regexp.MustCompile(`\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]`)
 // ESC [ <digits and semicolons only> m
 var sgrRe = regexp.MustCompile(`^\x1b\[[0-9;]*m$`)
 
-// sanitizeEscapes strips non-SGR terminal escape sequences from a line,
-// preventing untrusted content from injecting OSC/CSI/DCS control codes.
-// SGR sequences (colors/styles) are preserved.
+// sanitizeEscapes strips non-SGR terminal escape sequences and dangerous C0
+// control characters from a line, preventing untrusted content from injecting
+// OSC/CSI/DCS control codes or spoofing output via \r/\b overwrites.
+// SGR sequences (colors/styles), tabs, and newlines are preserved.
 func sanitizeEscapes(line string) string {
 	line = nonCSIEscRe.ReplaceAllString(line, "")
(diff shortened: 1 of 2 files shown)


wesm:
I think I'm at a stopping point here, the review findings are in the realm of esoterica. I'll merge this and release once the CI is green

wesm:
```
Docs mismatch for "all tests" command — AGENTS.md documents go test -tags integration ./... as   
  "(all)" which is accurate for the non-postgres test suite. Postgres tests require separate       
  infrastructure (a running database) and are intentionally behind a distinct -tags postgres flag.
  This distinction is by design, not an omission.

  Config-to-renderer tab width coverage gap — The clamping logic is 4 lines of if/else in
  newMarkdownCache and independently in truncateLongLines. Both are exercised by existing tests.
  Adding a test that verifies Go correctly passes an integer through a function call isn't
  valuable.

  Timeout integration coverage regression — TestWaitForAnalysisJob_Timeout still exists in
  analyze_test.go:321 and runs with unit tests (no build tag). It was never moved to the
  integration file because it's fast. Only the slow TestWaitForAnalysisJob (which blocks for real
  time) was tagged as integration.
```

wesm:
"C1 control characters are only interpreted as terminal commands in
  legacy 8-bit encoding modes (ISO 8859-1). All modern terminals in UTF-8 mode — which is the
  universal default — do not interpret U+0080–U+009F as control sequences. Since AI agent output is
   UTF-8 text, this is not a practical attack surface."
   
Merging!

PR #244, opened 2026-02-11
Chore skill tuneup

hughdbrown:
A collection of improvements to the SKILL.md files, i.e. the files that are core to the evaluation of reviewing and fixing code in this app.

Oddly, the changed SKILL.md files do not rate themselves as perfect. They laid a trap, inviting the developer to change the use of bash `&&` operators to make two SKILLs consistent, but afterwards noting that this change has the potential to re-order handling of `P` and `F` verdicts.


wesm:
We're going to need to tune the review prompts so that the agents don't get so worked up whenever we modify the skills definitions, so don't read too much into this feedback

PR #245, opened 2026-02-11
Add roborev config subcommand for get/set/list

mariusvniekerk:
## Summary
- Adds `roborev config` CLI command (similar to `git config`) for inspecting and modifying configuration without manually editing TOML files
- Supports `get`, `set`, and `list` subcommands with `--global`/`--local`/`--show-origin` flags
- Uses reflection-based key-value operations with dot notation for nested keys (e.g., `sync.enabled`, `ci.poll_interval`)
- Partial TOML writes via raw map manipulation to avoid polluting config files with default values
- Sensitive values (API keys, postgres URL) are masked in list output

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] `go vet ./...` passes
- [ ] Manual: `roborev config list --show-origin` shows defaults, global, and local provenance correctly
- [ ] Manual: `roborev config get default_agent` returns effective merged value
- [ ] Manual: `roborev config set default_agent claude-code --global` persists without polluting config with defaults
- [ ] Manual: `roborev config list --local` shows only repo-level overrides

### DIFF
--- a/cmd/roborev/config_cmd.go
+++ b/cmd/roborev/config_cmd.go
@@ -438,17 +438,26 @@ func atomicWriteConfig(path string, raw map[string]interface{}, isGlobal bool) e
 	if isGlobal {
 		dirMode = 0700
 	}
-	if err := os.MkdirAll(filepath.Dir(path), dirMode); err != nil {
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, dirMode); err != nil {
 		return err
 	}
+	// MkdirAll is a no-op for existing dirs. Tighten global config dir
+	// in case it was created with a permissive umask.
+	if isGlobal {
+		if err := os.Chmod(dir, dirMode); err != nil {
+			return err
+		}
+	}
 
-	// Preserve original file permissions if the file exists.
-	// Default to 0600 for global config (may contain secrets), 0644 for repo config.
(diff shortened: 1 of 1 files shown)


wesm:
Don't get too taken in by the security review paranoia!

wesm:
I grinded on this quite a bit. Barring blockers I'll merge this, and I opened a docs PR here https://github.com/roborev-dev/roborev-docs/pull/37

wesm:
"Not valid. The sensitive fields (anthropic_api_key, github_app_private_key) are defined on the
  global Config struct (line 81, 105), not on RepoConfig. The validateKeyForScope function
  validates against the struct definitions:                                                        
   
  - Non-global write → tries SetConfigValue on &RepoConfig{} first (line 409-410)                  
  - If the key doesn't exist in RepoConfig, it falls through to error                    

  Since RepoConfig has no sensitive fields (it's agent names, model strings, reasoning levels, CI
  review types — all non-secret configuration), a roborev config set ci.github_app_private_key
  value (without --global) would fail validation because that key doesn't exist in the RepoConfig
  struct. The scope enforcement is structural — if the field isn't in the struct, it can't be set.

  The finding incorrectly assumes RepoConfig "may include" sensitive fields like
  github_app_private_key. It doesn't — those are global-only by struct definition."

PR #260, opened 2026-02-14
fix: don't hardcode obsolete model for opencode agent

### LINKED ISSUE #259: OpenCode agent hardcodes obsolete MiniMax model
## Bug Description
The OpenCode agent hardcodes `opencode/minimax-m2.1-free` as the default model, which appears to be obsolete/unavailable. This causes failures when using the opencode agent without explicitly specifying a model.

## Location
`internal/agent/opencode.go` - lines 114 and 131

## Root Cause
The model is hardcoded in two places instead of being configurable or omitted entirely (letting OpenCode use its own default).

## Suggested Fix
Rather than hardcoding another model that may also become obsolete, the better approach is to **not pass `--model` at all when no model is explicitly configured**, allowing OpenCode to use its own default model (which it manages/updates).
This follows the principle of least surprise and avoids roborev needing to track OpenCode's model availability.

midnattsol:
- Remove hardcoded `opencode/minimax-m2.1-free` model which is obsolete
- Only pass `--model` flag when explicitly configured by user
- Let OpenCode use its own default model

Fixes #259

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -1,3 +1,4 @@
+review_agent = ""
 review_guidelines = """
 The daemon and client evolve in lockstep, and the daemon is restarted after
 updates and so API changes do not require backward compatibility shims, and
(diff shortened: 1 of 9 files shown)


wesm:
I tried pushing to your main branch but made a mess, I'll open a new PR I can control!

PR #278, opened 2026-02-16
Add compact command for verifying and consolidating reviews 

hughdbrown:
## Summary

Adds `roborev compact` — a command that consolidates multiple unaddressed review findings into a single verification pass. Instead of addressing each review individually, compact collects the outstanding findings and asks an agent to check which ones are still relevant against the current codebase.

## Usage

```bash
# Compact unaddressed reviews on the current branch (default: up to 20)
roborev compact

# Wait for the verification to finish and show results
roborev compact --wait

# Dry run — show which reviews would be compacted
roborev compact --dry-run

# Compact reviews across all branches
roborev compact --all-branches

# Customize agent, limit, or timeout
roborev compact --agent claude-code --limit 50 --timeout 15m
```

## How it works

1. Queries unaddressed review jobs for the current branch (or all branches with `--all-branches`)
2. Fetches the review outputs in batch
3. Builds a verification prompt containing all findings
4. Enqueues a compact job that runs an agent to check findings against the current code
5. On completion, marks the original source reviews as addressed

By default, compact enqueues the job and returns immediately. Use `--wait` to block until the agent finishes and display the consolidated result.

## What's included

- **Compact command** with `--dry-run`, `--wait`, `--limit`, `--timeout`, `--all-branches`, `--branch`, `--agent`, `--model`, `--reasoning` flags
- **`compact` job type** in the daemon/storage layer, with P/F verdict support in the TUI
- **Source job tracking** via sidecar metadata — compact jobs automatically mark their source reviews as addressed on completion
- **Output prefix** support so compact reviews include context (file list, source job IDs) prepended to the agent output
- **Batch job fetching** (`/api/jobs/batch`) to avoid N+1 API calls when collecting review outputs
- **`IsPromptJob()` method** to correctly distinguish prompt-native jobs (task, compact) from review jobs during retry

## Bug fixes included

- Fix negative unaddressed count in TUI when `hideAddressed` is active
- Fix prompt-selection regression where retried review jobs were misrouted through the stored-prompt path
- Fix `ClaimJob` not loading `output_prefix`, making worker-side validation a no-op
- Fix compact jobs being eligible as compaction sources (recursive self-compaction)
- Fix cancel race where canceled compact jobs could still mark source reviews as addressed

## Testing

```bash
# Run all tests
go test ./...

# Manual testing
roborev compact --dry-run          # verify it finds the right reviews
roborev compact --wait             # run a full compaction cycle
roborev status                     # check the compact job in the queue
```

PR #280, opened 2026-02-17
Preserve review history across rebases

wesm:
## Summary

Adds rebase-aware review preservation so reviews survive `git rebase` and `git commit --amend`.

- Store `patch_id` (content hash via `git patch-id --stable`) at enqueue time to match reviews across rebases
- Add `post-rewrite` git hook that calls `roborev remap` after rebase/amend to update job refs
- Add `/api/remap` daemon endpoint with proper error responses (400 for non-git paths, 404 for unknown repos, 500 for DB errors)
- Add `roborev remap` CLI command that reads old/new SHA pairs from stdin, filters by matching patch-id, and sends mappings to the daemon
- Hook install/uninstall handles both `post-commit` and `post-rewrite` with append/upgrade/skip behavior
- SQLite and PostgreSQL migrations add `patch_id` column with index (schema v5)
- Sync support: `patch_id` round-trips through push/pull
- Input validation: SHA format checks on remap requests, `repo_path` validated before git resolution
- `GetPatchID` uses `-c color.ui=false` to prevent user git config from affecting patch-id computation

## Test plan

- [x] `go test ./...` passes
- [x] `go test -tags integration ./internal/daemon/` — rebase and amend integration tests pass
- [x] `roborev init` installs post-rewrite hook alongside post-commit
- [x] `roborev install-hook` appends to existing non-roborev hooks, upgrades outdated ones
- [x] `roborev uninstall-hook` removes both post-commit and post-rewrite snippets
- [x] Make a commit, review it, rebase, verify `roborev show <new-sha>` finds the review
- [x] Same flow with `git commit --amend` (message-only change)
- [x] Remap against non-git path returns 400, unregistered repo returns 404

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/roborev/hook_test.go
+++ b/cmd/roborev/hook_test.go
@@ -805,6 +805,36 @@ func TestInstallOrUpgradeHook(t *testing.T) {
 		}
 	})
 
+	t.Run("refuses upgrade of non-shell hook mentioning roborev", func(t *testing.T) {
+		repo := testutil.NewTestRepo(t)
+		if err := os.MkdirAll(repo.HooksDir, 0755); err != nil {
+			t.Fatal(err)
+		}
+		hookPath := filepath.Join(repo.HooksDir, "post-commit")
+		// Non-shell hook that mentions roborev but has no version marker.
+		pythonHook := "#!/usr/bin/env python3\n# reviewed by roborev\nprint('hello')\n"
+		if err := os.WriteFile(hookPath, []byte(pythonHook), 0755); err != nil {
+			t.Fatal(err)
+		}
+
+		err := installOrUpgradeHook(
+			repo.HooksDir, "post-commit",
+			hookVersionMarker, generateHookContent, false,
+		)
+		if err == nil {
+			t.Fatal("expected error for non-shell hook in upgrade path")
+		}
+		if !strings.Contains(err.Error(), "non-shell interpreter") {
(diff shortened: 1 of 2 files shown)


wesm:
I will rebase this after #278 goes in which has better review guidelines which will get the security review to chill out

wesm:
"⏺ Summary of the 5 findings:                                                                        
                                                                                                    
  1. Unauthenticated remap endpoint — Not valid. Explicitly covered by review guidelines: daemon is 
  localhost-only, no auth needed.                                                                   
  2. Missing server-side SHA validation — Not valid. CLI validates with regex at the trust boundary;
   guidelines say don't re-flag at internal call sites.                                             
  3. Remap masks DB errors as "skipped" — Marginal validity. RemapJob returns (0, nil) for no-match
  and (0, error) for real failures, and real failures are silently swallowed. But the impact is low 
  — DB errors during remap would be systemic and visible elsewhere. Not worth changing the API
  contract for this edge case.
  4. CLI remap exits success on failure — Intentional. The return nil is explicitly commented "Don't
   fail the hook" — a non-zero exit from post-rewrite would print errors after every rebase.
  5. Pipe deadlock in GetPatchID — Valid, fixed. Closing the pipe reader when patchID.Start() fails
  prevents show.Wait() from hanging if the diff exceeds the pipe buffer."

PR #283, opened 2026-02-17
feat: add backup agent failover for review jobs

nstrayer:
## Human summary
I found myself hitting codex limits really quickly and then dumping back to copilot in this case. I wanted a way to automate this so I didnt have to keep switching the config.toml. This PR adds a `*backup_agent` setting as a fallback in case the review fails.

Coincidentally while making this PR I ran out of codex and the fallback worked perfectly!

## Summary

- Add per-workflow and default backup agent configuration (`default_backup_agent`, `review_backup_agent`, etc.) with the same repo-overrides-global priority as primary agents
- Implement failover in the worker pool: when an agent error exhausts retries, the worker resolves the backup agent from config and requeues the job with the new agent
- Distinguish agent errors (eligible for failover) from non-agent errors (prompt build failures) in the worker's retry logic
- No schema migration — backup agent is resolved from config at failover time, not stored per-job

## Changes

- `internal/config/config.go` — backup agent fields on `Config`/`RepoConfig`, `ResolveBackupAgentForWorkflow` resolution
- `internal/storage/jobs.go` — `FailoverJob(jobID, workerID, backupAgent)` atomically swaps agent and requeues
- `internal/daemon/worker.go` — `resolveBackupAgent` reads config at failover time, canonicalizes (verify installed, skip if same as primary), `failOrRetryAgent` path with failover split from `failOrRetry`
- Tests for config resolution, storage failover, worker-level backup agent resolution, and edge cases

## Test plan

- [x] `go test ./...` passes
- [ ] Verify failover works end-to-end: configure `backup_agent`, trigger a review with a failing primary agent, confirm job retries with the backup
- [ ] Verify no failover when `backup_agent` is unset (existing behavior unchanged)
- [ ] Verify second failover with same backup is a no-op (agent already swapped)

### DIFF
--- a/internal/daemon/worker_test.go
+++ b/internal/daemon/worker_test.go
@@ -378,6 +378,14 @@ func TestResolveBackupAgent(t *testing.T) {
 			want:       "test",
 		},
 		{
+			name:       "design review type uses design workflow",
+			jobAgent:   "codex",
+			reviewType: "design",
+			cfgField:   "DesignBackupAgent",
+			cfgValue:   "test",
+			want:       "test",
+		},
+		{
 			name:       "workflow mismatch returns empty",
 			jobAgent:   "codex",
 			reviewType: "security",
@@ -413,6 +421,7 @@ func TestResolveBackupAgent(t *testing.T) {
 			pool := NewWorkerPool(nil, NewStaticConfig(cfg), 1, NewBroadcaster(), nil)
 			job := &storage.ReviewJob{
 				Agent:      tt.jobAgent,
+				RepoPath:   t.TempDir(),
 				ReviewType: tt.reviewType,
 			}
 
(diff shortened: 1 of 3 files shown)


wesm:
ah, thanks for looking at this! I was planning to do this eventually since I was having to manually switch over to gemini after limiting out my ChatGPT Pro plan, but failing over gracefully is much better.

wesm:
Thanks @nstrayer — I'm going to rebase this and take a review pass on it

wesm:
The main feedback I have on this is that I want to pull the backup agent out of the database table and resolve it from the config at failover time — my assumption is that failover is going to be more exceptional for most users and so there is some possibility of things like:

1) user enqueues review
2) user modified config, changing failover agent
3) different failover agent is used for the review from step 1) than was intended at enqueue time

this strikes me as a bit contrived and won't really cause issues in practice. I'll push changes here shortly after I refine them a bit

wesm:
thanks @nstrayer!

PR #294, opened 2026-02-18
Add GitHub commit status checks, ref-aware config, and CI hardening

wesm:
## Summary

- **GitHub commit status API**: CI poller now sets commit status checks (`pending` → `success`/`failure`/`error`) on PR head commits via the GitHub Statuses API. Mixed-outcome batches (some jobs succeeded, some failed) get `failure`, not `success`. Uses `GitHubAppTokenProvider.APIRequest` for authenticated calls.
- **Ref-aware review guidelines**: `loadGuidelines` reads `.roborev.toml` from the default branch (`origin/main`, `origin/master`, etc.) instead of the working tree. Branch-specific guidelines are intentionally ignored to prevent prompt injection from untrusted PR authors. Filesystem fallback applies only when the default branch has no config.
- **CI poller config from default branch**: `loadCIRepoConfig` resolves review types, agents, and reasoning from the default branch post-fetch, so config changes take effect without rebuilding the daemon.
- **Commit SHA in review comments**: Synthesized, raw, and all-failed batch comments include the short HEAD SHA in the header for traceability.
- **Version in startup log**: Daemon logs `version.Version` at startup for easier debugging in systemd journal output.

## Test plan

- [x] `go test ./internal/daemon/...` — commit status transitions (pending, success, failure, error, mixed-outcome, comment post failure), `loadCIRepoConfig` (default branch, fallback, parse error propagation), comment formatting with SHA
- [x] `go test ./internal/config/...` — `LoadRepoConfigFromRef` (load from ref, nonexistent ref, file missing)
- [x] `go test ./internal/prompt/...` — `loadGuidelines` (non-main default branch, branch guidelines ignored, filesystem fallback, parse error blocks fallback), end-to-end single/range prompts with guidelines
- [x] `go build ./...`, `go vet ./...`, `go fmt ./...`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
@@ -13,6 +14,26 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/roborev-dev/roborev/internal/git"
 )
+
+// ConfigParseError is returned when .roborev.toml exists but
+// contains invalid TOML. Callers can check with errors.As.
+type ConfigParseError struct {
+	Ref string
+	Err error
+}
+
+func (e *ConfigParseError) Error() string {
+	return fmt.Sprintf("parse .roborev.toml at %s: %v", e.Ref, e.Err)
+}
+
+func (e *ConfigParseError) Unwrap() error { return e.Err }
+
+// IsConfigParseError reports whether err (or any error in its chain)
+// is a ConfigParseError.
+func IsConfigParseError(err error) bool {
+	var pe *ConfigParseError
+	return errors.As(err, &pe)
+}
 
 // HookConfig defines a hook that runs on review events
 type HookConfig struct {
@@ -476,7 +497,7 @@ func LoadRepoConfigFromRef(repoPath, ref string) (*RepoConfig, error) {
 
 	var cfg RepoConfig
 	if _, err := toml.Decode(string(data), &cfg); err != nil {
-		return nil, fmt.Errorf("parse .roborev.toml at %s: %w", ref, err)
+		return nil, &ConfigParseError{Ref: ref, Err: err}
 	}
 	return &cfg, nil
 }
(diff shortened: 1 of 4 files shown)


wesm:
sweet, status checks

image

wesm:
"All three are the same false positive from the previous review cycle, repackaged.
                                                                                                                
  Finding 1 (medium): Working-tree guideline fallback = prompt injection — No. The working tree is the operator's
  checkout, not attacker-controlled content. The daemon is read-only with respect to the working tree (CLAUDE.md:    
  "Daemon tasks must not modify the git working tree"). The CI poller fetches PR metadata via the GitHub API — it
  doesn't check out PR branches. There is no path from "untrusted PR author" to "controls working-tree               
  .roborev.toml." The fallback exists for the bootstrap case: repos that haven't committed .roborev.toml to their
  default branch yet, or repos without a remote.
                                               
  Finding 2 (medium): CI repo config fallback = branch-local policy override — Same incorrect threat model.
  loadCIRepoConfig reads operational settings (agents, review types, min_severity) from the operator's filesystem. A
  PR author cannot influence the operator's working tree. Additionally, this is operational config, not prompt
  content — even if the values differed, it only affects which agents run, not what instructions they receive.

  Finding 3 (medium): Invalid ref misclassified as "missing file" — Technically possible for pathological refs (e.g.,
   a ref manually pointed at a blob object), but GetDefaultBranch already validates refs with git rev-parse --verify
  --quiet before returning them, so LoadRepoConfigFromRef only receives validated refs in practice. The one scenario
  where the sentinel matching could misfire (blob ref) produces the correct behavior anyway — there's no tree to
  search, so "file doesn't exist" is functionally accurate, and the fallback to filesystem config is benign. Adding a
   pre-validation rev-parse call to LoadRepoConfigFromRef would add an extra subprocess per invocation to guard
  against a scenario that can't arise through normal git operations.

  Verdict: All three are invalid. They share a flawed threat model that assumes the working tree contains
  attacker-controlled content. It doesn't — roborev's daemon architecture prevents that."

PR #304, opened 2026-02-19
Test fix: resolve map aliasing and add stateful nav test

wesm:
This PR fixes a potential false-positive in TUI tests due to map aliasing and adds a new test case for stateful queue navigation.

### Changes
- **Fix Map Aliasing**: In `TestTUIPendingAddressedTransitions`, `initialPending` and `initialReviewPending` maps were being assigned by reference to the model. Since the model mutates these maps during updates (removing processed items), the subsequent assertions were iterating over modified maps, potentially skipping checks. This is fixed by deep-copying the maps using `maps.Copy` before assignment.
- **Add Stateful Navigation Test**: Added `TestTUIQueueNavigationSequences` to verify sequential keypress behavior in the queue view (e.g., moving down multiple times, then up, then jumping to top). This covers a gap left by the previous table-driven refactor which focused on single-step transitions.

### DIFF
--- a/cmd/roborev/tui_queue_test.go
+++ b/cmd/roborev/tui_queue_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"maps"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


wesm:
agent gone wild, rolled back

PR #312, opened 2026-02-20
feat: add 'roborev ci review' command and 'init gh-action' workflow generator

axsaucedo:
## Summary

- Adds `roborev init gh-action` — generates a GitHub Actions workflow file for running roborev reviews on pull requests, with multi-agent support, version pinning, and supply-chain hardening (SHA-pinned actions, checksum verification)
- Adds `roborev ci review` — daemon-free CLI command that runs batch reviews, synthesizes results, and optionally posts PR comments directly from CI
- Extracts `internal/review` package for batch execution, synthesis, and comment formatting, shared by both the daemon CI poller and the new CLI command
- Extracts `internal/ghaction` package for workflow generation with input validation, agent-aware env/secret wiring, and injection protection
- Converges `ci review` and CI poller behavior: workflow-aware agent/model resolution, shared constants, shared comment formatting

## Key design decisions

- `ci review` runs without a daemon — it invokes agents directly so the workflow doesn't need to manage daemon lifecycle
- `ci review` uses the same `config.ResolveAgentForWorkflow` / `ResolveModelForWorkflow` paths as the CI poller, so per-workflow config like `security_agent` works consistently in both modes
- Synthesis falls back to raw formatting when the synthesis agent is unavailable, so reviews always produce output
- `ci review` exits non-zero when all review jobs fail (quota-skipped batches exit 0)
- Invalid `--reasoning`, `--min-severity`, and `--review-types` flags return errors instead of silently falling back to defaults
- Copilot's GITHUB_TOKEN comes from the hardcoded `GH_TOKEN` line in the workflow template, not a separate secret entry
- Config load errors in `ci review` are logged as warnings, not hard failures — the command works with zero config
- SHA display standardized to 7 chars (`git.ShortSHA`) across all files, matching git's default

## Cleanup included in this branch

- Deduplicate `splitTrimmed` (ghaction.go had a copy)
- Consolidate `GITHUB_EVENT_PATH` parsing into shared `ghPREvent` struct
- Extract `review.MaxCommentLen` constant (was duplicated as local `60000` in 4 places)
- Add `config.ReviewTypeDefault/Security/Design` and `review.ResultDone/ResultFailed` constants, replace string literals in production code
- Standardize SHA shortening to `git.ShortSHA` (delete 2 duplicate implementations, replace ~30 inline `[:7]`/`[:8]` slices)

## Files

| Area | Files |
|------|-------|
| CLI commands | `cmd/roborev/ci.go`, `cmd/roborev/ghaction.go`, `cmd/roborev/main.go` |
| Review package | `internal/review/batch.go`, `internal/review/result.go`, `internal/review/synthesis.go`, `internal/review/synthesize.go` |
| Workflow generation | `internal/ghaction/ghaction.go` |
| Shared constants | `internal/config/config.go` (`ReviewType*`), `internal/review/result.go` (`ResultDone/Failed`, `MaxCommentLen`) |
| SHA utilities | `internal/git/git.go` (`ShortSHA`) |
| Daemon convergence | `internal/daemon/ci_poller.go`, `internal/daemon/server.go`, `internal/daemon/hooks.go` |
| Callers updated | `cmd/roborev/fix.go`, `compact.go`, `refine.go`, `tui.go`, `tui_handlers.go`, `internal/prompt/prompt.go`, `internal/agent/test_agent.go` |

## Test plan

- [x] `go test ./...` passes (unit tests for all new and modified packages)
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] Workflow-aware agent resolution test (`TestRunBatch_WorkflowAwareResolution`)
- [x] Workflow-aware model resolution test (`TestRunBatch_WorkflowModelResolution`)
- [ ] Manual: `roborev init gh-action` generates valid workflow YAML
- [ ] Manual: `roborev ci review --ref HEAD~1..HEAD` runs a review without daemon

### DIFF
--- a/internal/review/synthesize_test.go
+++ b/internal/review/synthesize_test.go
@@ -10,6 +10,33 @@ import (
 	"github.com/roborev-dev/roborev/internal/agent"
 )
 
+// failingSynthAgent always returns an error from Review,
+// used to force synthesis fallback deterministically.
+type failingSynthAgent struct{}
+
+func (a *failingSynthAgent) Name() string {
+	return "failing-synth"
+}
+func (a *failingSynthAgent) Review(
+	_ context.Context, _, _, _ string, _ io.Writer,
+) (string, error) {
+	return "", errors.New("agent exploded")
+}
+func (a *failingSynthAgent) WithReasoning(
+	_ agent.ReasoningLevel,
+) agent.Agent {
+	return a
+}
+func (a *failingSynthAgent) WithAgentic(_ bool) agent.Agent {
+	return a
+}
+func (a *failingSynthAgent) WithModel(_ string) agent.Agent {
+	return a
+}
+func (a *failingSynthAgent) CommandLine() string {
+	return "failing-synth"
+}
+
 // capturingAgent records the gitRef passed to Review.
 type capturingAgent struct {
 	capturedGitRef string
@@ -92,8 +119,11 @@ func TestSynthesize_SingleSuccess(t *testing.T) {
 }
 
 func TestSynthesize_MultipleResults_FallsBackToRaw(t *testing.T) {
-	// Without a real agent for synthesis, it should fall back
-	// to raw format.
+	// Register an agent that always fails so the test is
+	// deterministic regardless of what other agents exist
+	// in the global registry.
+	agent.Register(&failingSynthAgent{})
+
 	results := []ReviewResult{
 		{
 			Agent:      "codex",
@@ -111,7 +141,7 @@ func TestSynthesize_MultipleResults_FallsBackToRaw(t *testing.T) {
 
 	comment, err := Synthesize(
 		context.Background(), results, SynthesizeOpts{
-			Agent:   "nonexistent-synthesis-agent",
+			Agent:   "failing-synth",
 			HeadSHA: "def456789012",
 		})
 	if err != nil {


wesm:
I haven't had a chance to look at this closely yet but this should try to advantage of the existing CI configuration work in 

https://www.roborev.io/integrations/github/

Personally I wouldn't use this (instead preferring the CI poller that runs in this repository — this runs on a machine in my tailnet):

* I want to be able to use local LLMs for reviews in private repositories
* I want to be able to review using my spare Claude/Codex/Gemini subscription capacity, since the pay-as-you-go API use can get expensive 
* I want to maintain a persistent database of reviews run

I can see why this would be convenient for some people though!

wesm:
I checked out the branch and I'm working on this a bit

wesm:
"Thanks for the review. Assessed all five findings:

  Finding 1 (.npmrc injection) — Not applicable. This commit doesn't touch .npmrc handling. The
  generated workflow runs agent CLI tools, not npm packages.

  Finding 2 (Credential cross-exposure) — Not applicable. This commit adds GlobalConfig passthrough
  to BatchConfig for agent/model resolution. It doesn't change how credentials are scoped or passed.
   The existing trust model is: the daemon process has access to all configured API keys, same as
  before.

  Finding 3 (Unpinned agent installs) — Not applicable. This commit doesn't modify agent
  installation logic. The existing releases/latest pattern in ghaction.go already has a TODO for
  checksum pinning.

  Finding 4 (releases/latest redirect) — Not applicable. Same as above — no changes to the
  installation workflow in this commit.

  Finding 5 (Global CI reasoning fallback ignored) — Not valid. The global CIConfig struct
  intentionally has no Reasoning field — CI reasoning is a per-repo setting only
  (RepoCIConfig.Reasoning). The ci review command's resolveCIReasoning correctly checks: flag → repo
   config → hardcoded "thorough" default, matching the CI poller's resolution in processPR. There is
   no global CI reasoning to fall back to.

  Summary: 0 of 5 findings are valid for this commit. No changes needed."

wesm:
"Finding 1 (High): Unpinned agent installs — Pre-existing, not introduced by this PR. Already has a
   TODO. Valid concern, wrong PR.
                                                                                                    
  Finding 2: ROBOREV_VERSION from "latest" — Same as above. Pre-existing, not touched by this diff.
                                                                                                    
  Finding 3: Unbounded CI batch fan-out — Theoretical. The agent list comes from user-controlled
  config (their own .roborev.toml or --agent flag). A user configuring 50 agents in their own CI    
  config is self-inflicted. The goroutines are lightweight and bounded by the number of configured
  agents × review types, which in practice is single digits. Not a real concern.
                                               
  Finding 4: Global [ci].reasoning fallback ignored — Not valid. Already investigated and confirmed:
   CIConfig has no Reasoning field. The struct definition is at config.go:212-246. This is the third
   time this phantom finding has appeared.

  Finding 5: Repo-level synthesis agent fallback ignored — Not valid. Already investigated and
  confirmed: RepoCIConfig has no SynthesisAgent field. The struct definition is at
  config.go:342-354. Also the third time.

  Finding 6: init gh-action --agent doesn't pass --agent to ci review — Let me check this one.

⏺ Searched for 3 patterns, read 2 files (ctrl+o to expand)     

⏺ Finding 6 is observing that init gh-action --agent codex,gemini installs those agents in the      
  workflow but the generated roborev ci review command doesn't pass --agent codex,gemini. At
  runtime, ci review resolves agents via its own precedence chain (flag → repo config → global      
  config → auto-detect). So if the user has a .roborev.toml with [ci].agents, it works. If they
  don't, it falls back to auto-detect, which will find the agents that were installed in the        
  workflow. The auto-detect fallback in resolveAgentList returns []string{""} (empty string =
  auto-detect), and GetAvailable("") picks the first available agent — it doesn't run all installed 
  agents.                                                                                           

  So this finding is partially valid: if you do init gh-action --agent codex,gemini, the workflow
  installs both but ci review without --agent will only auto-detect one. However, this is
  pre-existing behavior from the init gh-action feature, not introduced by this cleanup PR.

  ---
  Summary: 0 of 6 findings are introduced by this PR's diff. Findings 4 and 5 are factually wrong
  (referencing struct fields that don't exist) — this is the third time they've appeared. Findings
  1, 2, 3, and 6 describe pre-existing behavior untouched by the cleanup changes. None should block
  merge."

wesm:
thank you! I haven't tested out the gh action end to end but if you run into any issues just open an issue or a PR to fix!

axsaucedo:
Awesome! Thank you @wesm - testing as we speak!

PR #315, opened 2026-02-22
Store verdict for later use

hughdbrown:
Add verdict_bool to the reviews table, populated by ParseVerdict() at CompleteJob() time.


wesm:
"Not valid. This is the same phantom finding from #6595 repeated verbatim. There are no
  run/analyze/custom task jobs in the system. The valid review types are default, security, and
  design — all produce review output with verdict semantics. The finding is based on functionality
  that doesn't exist."

wesm:
I'm going to fix the test race failure that is unrelated to this PR in a new PR

PR #317, opened 2026-02-22
Add daemon activity log for operational visibility

wesm:
## Summary

- Add structured `ActivityLog` (JSONL file + 500-entry in-memory ring buffer) capturing routine daemon events: start/stop, zombie cleanup, job enqueue/start/complete/fail, and config reloads
- Expose recent entries via `GET /api/activity` with configurable limit
- Deep-copy details maps to prevent shared mutable state; guard against negative `RecentN` input; truncate log file at 5MB on open to prevent unbounded growth

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all existing + new tests)
- [x] New unit tests for ActivityLog: write/recent ordering, ring buffer eviction, JSONL file format, details deep-copy, negative RecentN, file truncation
- [x] New handler tests for `/api/activity`: method restriction, default/custom/clamped/invalid limits, nil log
- [ ] Manual: start daemon, submit a review, verify `~/.roborev/activity.log` contains `daemon.started` and `job.enqueued` entries
- [ ] Manual: `curl http://127.0.0.1:7373/api/activity` returns recent entries

### DIFF
--- a/internal/daemon/activitylog.go
+++ b/internal/daemon/activitylog.go
@@ -157,8 +157,8 @@ const rotateCheckInterval = 1000
 
 // maybeRotate checks the log file size every rotateCheckInterval
 // writes and truncates it to zero if over maxActivityLogSize. Must
-// be called with a.mu held. Truncates in-place to avoid
-// close/reopen failure modes.
+// be called with a.mu held. Closes and reopens the file because
+// Windows does not allow Truncate on an O_APPEND handle.
 func (a *ActivityLog) maybeRotate() {
 	a.writeCount++
 	if a.writeCount < rotateCheckInterval {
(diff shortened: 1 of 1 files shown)


wesm:
"Not valid. This is the same finding from review #6590 that I already rejected. The reasoning
  hasn't changed:
                                                                                                    
  1. git config --local --get with -C <path> exits 128 only when the path isn't a valid git repo.
  There's no realistic "other fatal 128" scenario for this specific command.                        
  2. cloneNeedsReplace already calls isValidGitRepo before cloneRemoteMatches, so the 128 path only
  fires in direct calls (like tests).        
  3. Returning an error for 128 would block re-clone behavior for broken directories, which is the
  opposite of what we want."

Issue #323, opened 2026-02-23
Failed to run fix "F" on compacted review (x10) in tui


danielaborg:
Error:

```
Job #76 failed: create worktree: ref must not be empty
```

wesm:
cc @mariusvniekerk

mariusvniekerk:
Taking a look, seems not too hard.

wesm:
Merged this PR, please let us know if the issue is resolved

biscuitvicious:
> Merged this PR, please let us know if the issue is resolved

@wesm There actually turned out to be some quite serious issues with the functionality mostly to do with disappearing git worktrees in /tmp, see my comments here: https://github.com/roborev-dev/roborev/pull/327#issuecomment-3947916479

Would you like me to raisse separate issues for these problems?

wesm:
> Would you like me to raisse separate issues for these problems?

yes

wesm:
after the next release please let us know if this issue persists

biscuitvicious:
> > Would you like me to raisse separate issues for these problems?
> 
> yes

@wesm Looks like you've made some improvements that may potentially fix some or all of my issues, let me test and I'll check if any problems persist in my workflow, in which case I will raise new issues. Thanks.

PR #327, opened 2026-02-23
fix: handle empty git ref when fixing compact review jobs

mariusvniekerk:
I think this likely gets us to  a fix for #323 

## Summary

- Compact jobs are stored with an empty `git_ref` (only a display label is stored). Attempting to fix a compact review would pass `""` to `worktree.Create`, causing `create worktree: ref must not be empty`.
- The fix job handler now resolves the ref via: explicit request override → parent job's `git_ref` → parent job's branch → `HEAD` (with a log warning).
- When the TUI detects no git ref and no branch (the HEAD fallback case), it shows a new confirmation modal (`tuiViewFixGitRef`) pre-filled with `"HEAD"` so the user can verify or correct the target before proceeding.

## Test plan

- [ ] Fix a compact review job from the TUI — should no longer fail with `create worktree: ref must not be empty`
- [ ] When compact job has a branch set, fix proceeds normally using that branch as the ref
- [ ] When compact job has neither git ref nor branch, the new git ref modal appears pre-filled with `HEAD`, allowing the user to edit before confirming
- [ ] `go test ./...` passes

### DIFF
--- a/internal/daemon/server_test.go
+++ b/internal/daemon/server_test.go
@@ -3565,6 +3565,7 @@ func TestHandleListJobsJobTypeFilter(t *testing.T) {
 
 func TestHandleFixJobStaleValidation(t *testing.T) {
 	server, db, tmpDir := newTestServer(t)
+	server.configWatcher.Config().FixAgent = "test"
 
 	repoDir := filepath.Join(tmpDir, "repo-fix-val")
 	testutil.InitTestGitRepo(t, repoDir)


biscuitvicious:
Can confirm with this PR branch the feature now works fine on my other compacted review (and did trigger the ref query).

biscuitvicious:
> Can confirm with this PR branch the feature now works fine on my other compacted review (and did trigger the ref query).

Actually I take that back, there were 2 issues that surfaced when I tried to get the review fixed:

1) The first time I entered HEAD as the target (should have picked the branch), and the agent got a bit lost in the /tmp worktree and ended up just *staging* the review changes and not committing them in the end, leading to their removal when the worktree was dropped.

2) Trying to redo the fix I ran the process again on the compacted review and entered a custom instruction to fix all review findings and commit them to the target branch. What happened then was that the review context seems to have got lost in the context of the fix job so the agent just didn't know what to do and ended up asking me what I'm talking about, at which point I had to kill it.

For the 2:nd and now 3:rd attempt, the compacted review appears to have picked the target branch automatically instead of asking what ref to commit to like it did the first time. I guess this is good but why couldn't it do it from the start?

Ok, I ran the fix now a 3rd time and the behaviour is the same as the first time - changes are staged and not committed and then dropped and lost.

@mariusvniekerk Do you want me to raise new issues about this? It seems pretty related to this PR so it seems unnecessary.

biscuitvicious:
I would prefer if I could 1) ensure the fix worktree is persisted after the agent completes its work, for exactly this kind of situation and 2) specify where the worktree is created, e.g. for example as a subdirectory of the project directory itself.

biscuitvicious:
Can confirm that the behaviour of not committing on the worktree and losing the changes also hppens on normal reviews. OpenCode.

wesm:
> @mariusvniekerk Do you want me to raise new issues about this? It seems pretty related to this PR so it seems unnecessary.

please raise a new issue about this so we can look at it

PR #328, opened 2026-02-23
feat: inline fix panel in review view

mariusvniekerk:
## Summary

- Replaces the full-screen `tuiViewFixPrompt` modal with an inline split-pane fix panel at the bottom of the review view
- Pressing `F` from the queue fetches the review and opens it with the fix panel focused; pressing `F` from the review view opens the panel inline
- `Tab` toggles keyboard focus between the review content (scroll mode) and the fix panel (input mode); a visible border color change indicates which panel is active
- `Enter` submits the fix and navigates to the tasks view; `Esc` closes the panel
- Uses `lipgloss.NormalBorder()` for the panel frame (eliminates manual box-drawing arithmetic)
- Removes `tuiViewFixPrompt`, `renderFixPromptView()`, `handleFixPromptKey()`, and `fixPromptFromView`

## Test Plan

- [ ] Press `F` from queue view — should navigate to review with fix panel open and focused (magenta border)
- [ ] Press `F` from review view — fix panel appears at bottom, focused
- [ ] `Tab` unfocuses fix panel (gray border), review scroll keys work; `Tab` again refocuses
- [ ] Type in fix panel when focused — text appears with cursor
- [ ] `Esc` when focused closes panel; `Esc` when unfocused also closes panel (stays in review)
- [ ] `Enter` submits and navigates to tasks view
- [ ] `go test ./...` passes (2970 tests)

### DIFF
--- a/cmd/roborev/tui_handlers.go
+++ b/cmd/roborev/tui_handlers.go
@@ -1140,12 +1140,10 @@ func (m tuiModel) handleEscKey() (tea.Model, tea.Cmd) {
 	} else if m.currentView == tuiViewReview {
 		// If fix panel is open (unfocused), esc closes it rather than leaving the review
 		if m.reviewFixPanelOpen {
-			m.reviewFixPanelOpen = false
-			m.reviewFixPanelFocused = false
-			m.fixPromptText = ""
-			m.fixPromptJobID = 0
+			m.closeFixPanel()
 			return m, nil
 		}
+		m.closeFixPanel()
 		returnTo := m.reviewFromView
 		if returnTo == 0 {
 			returnTo = tuiViewQueue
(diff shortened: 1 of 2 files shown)


wesm:
"This is the same invalid finding that's now appeared in four separate reviews. The server-side handleFixJob at
  internal/daemon/server.go:1963-1977 has a complete fallback chain:
                                                                                                                    
  1. Client-provided git_ref (if any)
  2. parentJob.GitRef (skipped if it's a range like sha..sha)                                                       
  3. parentJob.Branch                  
  4. "HEAD" (with a log message)            
                                               
  The old tuiViewFixGitRef modal was prompting the user to manually type a git ref, which was unnecessary friction
  -- the server resolves it correctly. The test suite at server_test.go:3783+ explicitly validates the branch
  fallback for branchless jobs and range-ref handling.

  Nothing to do here."

PR #329, opened 2026-02-23
feat: CI repo wildcard patterns and exclusion lists

dannysteenman:
## Summary

The CI poller's `[ci] repos` field only accepts explicit `"owner/repo"` entries today. For orgs with many repos this is tedious to maintain. This PR adds:

- **Glob pattern support** in `repos` (e.g., `"myorg/*"`, `"myorg/api-*"`) using Go's `path.Match` syntax. The owner part must be literal — no wildcards before the `/`.
- **`exclude_repos`** — glob patterns that filter the final resolved list (applies to both exact and expanded entries).
- **`max_repos`** — safety cap on total expanded repos (default: 100).
- Wildcard entries are expanded via `gh repo list <owner> --no-archived` with results cached for 1 hour (fixed internally, not user-configurable).
- Exact entries (`owner/repo`) continue working unchanged — no API calls needed.
- On `gh repo list` failure: logs a warning, skips that owner's wildcards, exact entries still pass through.

### Example config

```toml
[ci]
repos = ["myorg/*", "myorg/api-*", "other-owner/specific-repo"]
exclude_repos = ["myorg/archived-*", "myorg/internal-*"]
max_repos = 100
```

## Files changed

| File | Change |
|------|--------|
| `internal/config/config.go` | Added `ExcludeRepos`, `MaxRepos` fields to `CIConfig` + `ResolvedMaxRepos()` helper |
| `internal/daemon/ci_repo_resolver.go` | New file — `RepoResolver` with cache, wildcard expansion, exclusions, `max_repos` enforcement, `listReposFn` test seam |
| `internal/daemon/ci_poller.go` | Wired `RepoResolver` into poll loop with fallback to exact entries on error |
| `internal/config/config_test.go` | Tests for new config fields, defaults, and legacy `repo_refresh_interval` backward compat |
| `internal/daemon/ci_repo_resolver_test.go` | 16 tests: exact-only (zero API calls), wildcard expansion, exclusions, `max_repos` truncation, cache hit/invalidation, API failure fallback, deduplication, empty result caching |

## Design decisions worth discussing

1. **Refresh interval is not user-configurable** — hardcoded to 1 hour. The reasoning: one less knob for users to misconfigure, and 1h aligns with the typical CI poll cadence. The `BurntSushi/toml` library silently ignores unknown keys, so anyone with `repo_refresh_interval` in their existing config won't see errors (verified with a regression test). Would you prefer exposing this as a config option?

2. **Cache key includes `max_repos`** — changing `max_repos` invalidates the cache and triggers re-expansion. Alternative: always re-apply the cap on cache hits without re-querying. Current approach is simpler.

3. **Deduplication is case-insensitive** — `"Acme/Api"` and `"acme/api"` are treated as the same repo (GitHub is case-insensitive for owner/repo). The original casing from the first occurrence is preserved.

## Test plan

- [ ] `go test ./...` passes (all 16 new + existing tests)
- [ ] `go vet ./...` clean
- [ ] Manual: set `repos = ["<your-org>/*"]` in config, run daemon, confirm it discovers and polls expected repos
- [ ] Manual: verify `exclude_repos` filters correctly
- [ ] Manual: verify exact-only config makes zero GitHub API calls

### DIFF
--- a/internal/daemon/ci_repo_resolver_test.go
+++ b/internal/daemon/ci_repo_resolver_test.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


wesm:
very useful, thank you!

PR #336, opened 2026-02-24
fix: resolve worktree path when applying fix patches

mariusvniekerk:
## Summary

- When applying a fix patch, resolve the target directory by checking if the branch is checked out in an existing git worktree (via `git worktree list --porcelain`). Previously patches were always applied to the main repo path, which was wrong when the target branch lived in a separate worktree.
- If the branch is not checked out anywhere, show a confirmation modal prompting the user to create a temporary worktree, apply and commit there, then clean up. The commit persists on the branch.
- Added `WorktreePathForBranch` to `internal/git` with tests covering worktree lookup, no-worktree fallback, empty branch, and main repo HEAD cases.

Addresses feedback from @danielaborg on #327 (related to #323).

## Test plan

- [ ] Apply a fix patch when the target branch is checked out in a worktree — patch should be applied in the worktree directory
- [ ] Apply a fix patch when the target branch is checked out as HEAD in the main repo — works as before
- [ ] Apply a fix patch when the target branch is not checked out anywhere — confirmation modal appears, creating a temp worktree on confirm
- [ ] Cancel the worktree confirmation with esc/n — returns to tasks view without applying
- [ ] `go test ./...` passes (2965 tests)

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -686,18 +686,42 @@ func WorktreePathForBranch(repoPath, branch string) (string, bool, error) {
 	//   worktree /path/to/dir
 	//   HEAD <sha>
 	//   branch refs/heads/<name>
+	//   [prunable ...]
 	//   <blank line>
-	var currentPath string
+	//
+	// We collect path+branch pairs, then verify the path exists before
+	// returning it. This avoids returning stale/prunable worktree paths.
+	type wtEntry struct {
+		path, branch string
+	}
+	var entries []wtEntry
+	var currentPath, currentBranch string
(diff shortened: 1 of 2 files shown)


wesm:
this one is wily! I think I'm close to the finish line though

mariusvniekerk:
worktrees certainly make git a lot more interesting than assuming things are simple.

mariusvniekerk:
adding a test to assert the false claim is false, and added the other one as a test case which found the problem

wesm:
Merging this, the failure is due to a race condition (unrelated to this PR)

PR #349, opened 2026-02-25
Print right version when installed via "go install"

tebeka:
Currently when install with `go install github.com/roborev-dev/roborev/cmd/roborev@latest`, running `roborev version` prints `dev`. This PR changes that to take version info from debug info.


tebeka:
CI failure doesn't seem related.

Issue #316, opened 2026-02-22
feat: Agent Client Protocol (ACP)


VeitSanner:
Currently, for every supported agent the specific agent has to be hardcoded. The Agent Client Protocol allows to support different agents by configuration. 

Are there any plans to support ACP?

wesm:
I would be happy to look at a pull request. It’s important for things to “just work” for most users of roborev without requiring a bunch of configuration, but if ACP provides an escape hatch for advanced users that sounds fine to me.

veitsanner:
Thanks for your openness to considering ACP support! From what I’ve gathered, implementing ACP would essentially mean adding another agent interface implementation. The main challenge I foresee is integrating ACP’s more dynamic agent configuration with the existing set up in `config` and registry in the package `agent` - this might require some deeper adjustments. And of course, the existing setup should work as it is.

I’ll take a closer look at the required changes to get a better sense of the scope, but I’ll need to check if I can carve out the time to tackle this properly. 

Before diving in, I wanted to ask: Are there any development guidelines or best practices beyond what’s documented on [Development | roborev](https://www.roborev.io/development/) that I should be aware of?

wesm:
Nope. Just submit a pull request

veitsanner:
I found some time to look into this more closely. The current version is available in my fork [VeitSanner/roborev at feature/add-acp-as-separate-agent](https://github.com/VeitSanner/roborev/tree/feature/add-acp-as-separate-agent).

# Approach

## ACP Agent
- Add with `ACPAgent` a new implementation of the Agent interface.
- `ACPAgent` acts like a facade behind which different agent CLIs can be configured.
- `ACPAgent` uses [acp-go-sdk](https://github.com/coder/acp-go-sdk) to communicate with the agent's CLI.

## Configuration
- Because `ACPAgent` acts just as a facade, the actual agent CLI has to be configured in `roborev.toml`.
- Added confgiguration field `ACP` which holds the ACP agent config.

# Questions

- What is the preferred approach to provide the configuration when the ACPAgent is registered using the `ìnit()`? Or are there other options to create the ACPAgent with the info from the configuration?
- Would it make sense to have multiple ACPAgent configurations available in `roborev.tonl`? Since the ACPAgent acts as a facade this would then allow to support different agent CLIs for ACP as well. A specific agent configuration could be selected using `--agent=acp:copilot`. The segment after the colon needs to match the agent name in `roborev.toml`

wesm:
[...]

PR #354, opened 2026-02-25
Add support for Agent Client Protocol to support various agents

VeitSanner:
# Approach

## ACP Agent
- Add with `ACPAgent` a new implementation of the Agent interface.
- `ACPAgent` acts like a facade behind which different agent CLIs can be configured.
- `ACPAgent` uses [acp-go-sdk](https://github.com/coder/acp-go-sdk) to communicate with the agent's CLI.

## Configuration
- Because `ACPAgent` acts just as a facade, the actual agent CLI has to be configured in `roborev.toml`.
- Added confgiguration field `ACP` which holds the ACP agent config.

# Questions

- What is the preferred approach to provide the configuration when the ACPAgent is registered using the `ìnit()`? Or are there other options to create the ACPAgent with the info from the configuration?
- Would it make sense to have multiple ACPAgent configurations available in `roborev.tonl`? Since the ACPAgent acts as a facade this would then allow to support different agent CLIs for ACP as well. A specific agent configuration could be selected using `--agent=acp:copilot`. The segment after the colon needs to match the agent name in `roborev.toml`

Implements #316

### DIFF
--- a/internal/agent/acp.go
+++ b/internal/agent/acp.go
@@ -1342,12 +1342,8 @@ func (c *acpClient) WaitForTerminalExit(ctx context.Context, params acp.WaitForT
 	terminal, exists := c.getTerminal(params.TerminalId)
 
 	if !exists {
-		// Terminal not found - it may have already been cleaned up
-		// Return a default response indicating the terminal has exited
-		return acp.WaitForTerminalExitResponse{
-			ExitCode: nil, // nil indicates unknown exit status
-			Signal:   nil,
-		}, nil
+		return acp.WaitForTerminalExitResponse{},
+			fmt.Errorf("terminal %s not found", params.TerminalId)
 	}
 
 	// Wait for the command to finish


wesm:
I'm going to take a pass at this, so please hold off on pushing any more changes

wesm:
Some candid feedback: this project has a pretty strong security posture, so if you're going to use an open weights model like devstral-2, you need a stronger agent (i.e. codex-5.3 or gemini-3.1-pro) reviewing to find the security problems. I have those models reviewing PRs so you can get the feedback that way, but it's a great deal more productive to have the feedback in your local development loop. Likely the open weights models will improve so maybe in 12 or 24 months things will be different.

Anyway I am working on whipping this into shape now!

wesm:
Thank you!!

veitsanner:
> Some candid feedback: this project has a pretty strong security posture, so if you're going to use an open weights model like devstral-2, you need a stronger agent (i.e. codex-5.3 or gemini-3.1-pro) reviewing to find the security problems. I have those models reviewing PRs so you can get the feedback that way, but it's a great deal more productive to have the feedback in your local development loop. Likely the open weights models will improve so maybe in 12 or 24 months things will be different.
> 
> Anyway I am working on whipping this into shape now!

@wesm, many thanks for integrating my PR with the current code base. Very much appreciated.

And yes, I ran local checks with Mistral AIs devstral, but the feedback I got was not as good as the feedback from the models, roborev uses to review the PRs. Will take that into consideration.

PR #367, opened 2026-02-26
feat: handle managed daemon lifecycle on update

### LINKED ISSUE #366: roborev update conflicts with launchd-managed daemon
## Problem

When the daemon is managed by an external service manager (e.g. macOS launchd), `roborev update` blindly restarts the daemon after installing a new binary. This conflicts with the already-running launchd-managed instance:

```
❯ roborev update
...
Updated to v0.35.0
Restarting daemon... OK
```

```
2026/02/24 10:06:14 main.go:675: Starting roborev daemon (version v0.35.0)...
2026/02/24 10:06:14 main.go:771: Received signal terminated, shutting down...
2026/02/24 10:06:24 main.go:675: Starting roborev daemon (version v0.35.0)...
Error: daemon already running (pid 61213 on 127.0.0.1:7373)
```

The sequence is:
1. `roborev update` installs the new binary
2. It starts a new daemon process
3. launchd detects the old daemon died and restarts it (with the new binary)
4. The daemon started by `roborev update` fails with "daemon already running"

## Use case

Running `roborev ci` as a persistent daemon on a Mac Mini via launchd, so the daemon survives reboots without manual intervention. The user disabled the default `roborev daemon` auto-start and relies entirely on launchd for lifecycle management.

## Suggested behavior

- `roborev update` should detect if the daemon is managed by an external service manager and skip the restart step
- Alternatively, add a `--no-restart` flag to `roborev update` so users with external daemon management can opt out
- When the daemon is externally managed, the update could signal the running daemon to gracefully restart itself (e.g. via SIGHUP) instead of spawning a new process

Reported by @dsteenman.

wesm:
## Summary

- Add `--no-restart` flag to `roborev update` for daemons managed by external service managers (launchd, systemd)
- Detect when an external manager restarts the daemon after the old process is stopped, avoiding a duplicate start that conflicts with the managed instance
- Fall back to runtime file presence when the daemon health probe fails, so updates don't silently skip restart for a running-but-unresponsive daemon
- Replace `waitForDaemonPIDChange` with `waitForDaemonExit` that returns early as soon as the old process exits, eliminating the unconditional 2-second delay for unmanaged daemons

Closes #366

## Test plan

- [ ] Run `roborev update --no-restart` with a running daemon — daemon should remain running, "Skipping daemon restart" message printed
- [ ] Run `roborev update` with a launchd-managed daemon — should detect the manager-restarted PID and print OK without spawning a duplicate
- [ ] Run `roborev update` with a self-managed daemon — should stop, exit quickly (no 2s delay), start new binary, print OK
- [ ] Run `roborev update` when daemon is unresponsive but runtime files exist — should attempt stop+start instead of silently skipping
- [ ] Run `go test ./cmd/roborev/...` — all restart path tests pass

### DIFF
--- a/cmd/roborev/processcheck_unix.go
+++ b/cmd/roborev/processcheck_unix.go
@@ -49,7 +49,8 @@ func identifyPIDForUpdate(pid int) updatePIDIdentity {
 		return updatePIDNotRoborev
 	}
 
-	cmd := exec.Command("ps", "-p", strconv.Itoa(pid), "-o", "command=")
+	// Use -ww to request untruncated command output on BSD/macOS ps.
+	cmd := exec.Command("ps", "-ww", "-p", strconv.Itoa(pid), "-o", "command=")
 	output, err := cmd.Output()
 	if err != nil {
 		return updatePIDUnknown


wesm:
The implementer doesn't agree with these findings.

wesm:
acceptable risk. merging

Issue #405, opened 2026-03-01
Add fallback synthesis agent, or print uncollapsed CI reviews when synthesis agent is quota'd out


wesm:
The behavior when the synthesis agent is rate-limited is not great, so we can improve this

Issue #404, opened 2026-03-01
Add more granular CI review matrix capability


wesm:
Currently you can run multiple review types and multiple agents, but you can't run multiple review types with different agents for each review. Suggest evolving to support also a config.toml syntax like

```
[ci]
enabled = true
poll_interval = "1m"
repos = [...]
synthesis_agent = "codex"
min_severity = "medium"

[ci.reviews]
codex = ["security", "review"]
gemini = ["review"]
```

wesm:
this was done in #406

PR #406, opened 2026-03-02
Add CI review matrix, PR throttling, closed-PR cancellation, and expanded raw fallback

wesm:
## Summary

- **Granular review matrix** (#404): Add `[ci.reviews]` map for assigning specific review types to specific agents (e.g., `codex = ["security"]`, `gemini = ["security", "review"]`). Falls back to existing `agents × review_types` cross-product when not set. Map keys are sorted for deterministic job order.
- **PR throttling**: Add `throttle_interval` config (default `"1h"`) to prevent excessive reviews when contributors push frequently. Deferred PRs show a pending GitHub status with the next eligible review time. Set to `"0"` to disable.
- **Throttle bypass for specific users**: Add `throttle_bypass_users` config (default: `["wesm", "mariusvniekerk"]`) so listed maintainers bypass the throttle interval and get immediate reviews on every push. Matching is case-insensitive. PR author login is fetched via `gh pr list --json author`.
- **Expanded raw fallback** (#405): Replace collapsed `<details>` blocks in synthesis-unavailable comments with inline `###` headings and `---` separators so reviews are visible without clicking.
- **Synthesis backup agent** (#405): When the primary synthesis agent fails (quota, unavailable), try a configured `synthesis_backup_agent` before falling back to raw formatting.
- **Cancel batches for closed/merged PRs**: Fix infinite retry loop where `postBatchResults` failures cause batches to retry forever via `reconcileStaleBatches`. Two-layer fix: `pollRepo` proactively cancels batches for closed PRs each poll cycle (with `isPROpen` verification to avoid false cancellation when `gh pr list` truncates at 100), and `postBatchResults` checks PR state after claiming a batch and finalizes it if closed.
- **Repo-level review disabling**: Repo `[ci.reviews]` with empty lists now correctly disables reviews for that repo instead of falling through to the global matrix.
- **Empty matrix guard**: Skip batch creation when the review matrix is empty to avoid stuck zero-job batches. Superseded-batch cancellation still runs before this guard.
- **Claimed-batch safety**: `CancelClosedPRBatches` skips batches with `claimed_at IS NOT NULL` to avoid racing with synthesis/posting.
- **Error status on agent failure**: Set an error commit status when agent resolution fails during batch setup (quota, misconfiguration) so PR authors get immediate feedback instead of silent retries.
- **Context threading**: `isPROpen` checks in `pollRepo` use the poll context so they respect cancellation/shutdown instead of blocking on `context.Background()`.

## Test plan

- [x] `go build ./...` and `go vet ./...` pass
- [x] `go test ./internal/config/...` — review matrix, throttle interval parsing, TOML field parsing, `IsThrottleBypassed` case-insensitive matching, `RepoCIConfig.ResolvedReviewMatrix` empty/nil semantics
- [x] `go test ./internal/storage/...` — `LatestBatchTimeForPR`, `GetPendingBatchPRs`, `CancelClosedPRBatches` (including claimed-batch skip)
- [x] `go test ./internal/review/...` — updated raw batch comment format
- [x] `go test ./internal/daemon/...` — throttle behavior, throttle bypass, reviews map matrix, repo-level reviews override, repo empty reviews disables jobs, closed-PR batch cancellation with `isPROpen` verification, skip cancel when PR still open, empty matrix skips batch, empty matrix still cancels superseded, agent failure sets error commit status, `isPROpen` guard in `postBatchResults`, partial-cancel worker kill, backup synthesis agent

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -348,8 +348,14 @@ func (c *CIConfig) ResolvedReviewMatrix() []AgentReviewType {
 // the cross-product of Agents x ReviewTypes (which may be empty,
 // meaning "use global").
 func (c *RepoCIConfig) ResolvedReviewMatrix() []AgentReviewType {
-	if len(c.Reviews) > 0 {
-		return reviewsMapToMatrix(c.Reviews)
+	if c.Reviews != nil {
+		// Reviews map is configured — return the resolved matrix
+		// even when empty (signals "disable reviews for this repo").
+		m := reviewsMapToMatrix(c.Reviews)
+		if m == nil {
+			return []AgentReviewType{}
+		}
+		return m
 	}
 	return nil
 }
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1911,6 +1911,32 @@ func TestRepoCIConfigResolvedReviewMatrix(t *testing.T) {
 			t.Errorf("got %v", matrix[0])
 		}
 	})
+
+	t.Run("empty Reviews disables reviews", func(t *testing.T) {
+		ci := RepoCIConfig{
+			Reviews: map[string][]string{},
+		}
+		matrix := ci.ResolvedReviewMatrix()
+		if matrix == nil {
+			t.Fatal("expected non-nil empty slice, got nil")
+		}
+		if len(matrix) != 0 {
+			t.Errorf("expected 0 entries, got %d", len(matrix))
+		}
+	})
+
+	t.Run("Reviews with all empty lists disables reviews", func(t *testing.T) {
+		ci := RepoCIConfig{
+			Reviews: map[string][]string{"codex": {}},
+		}
+		matrix := ci.ResolvedReviewMatrix()
+		if matrix == nil {
+			t.Fatal("expected non-nil empty slice, got nil")
+		}
+		if len(matrix) != 0 {
+			t.Errorf("expected 0 entries, got %d", len(matrix))
+		}
+	})
 }
 
 func TestResolvedThrottleInterval(t *testing.T) {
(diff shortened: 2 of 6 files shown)


wesm:
"Read 2 files (ctrl+o to expand)                                                                               
   
⏺ All four findings are invalid. Here's why:                                                                    
                                         
  Finding 1 (High): "PR throttle check overwrites commit status" — Invalid. Lines 290-296 show                  
  HasCIBatch(ghRepo, pr.Number, pr.HeadRefOid) runs before the throttle check at line 307. Once a batch exists
  for the current HEAD, processPR returns at line 295 and never reaches the throttle logic. The finding's       
  suggested fix ("Add a HasCIBatch check at the very beginning") is recommending something that already exists.
  This is the same incorrect finding from the previous review round.
                                               
  Finding 2 (Medium): "Throttle early-return prevents superseded-batch cleanup" — Invalid. This was fixed in the
   previous commit (b2712217). CancelSupersededBatches now runs at line 443, well after the throttle check — but
   the throttle check only fires for different SHAs. When a new commit is pushed, the throttle checks
  LatestBatchTimeForPR (which looks at any SHA for this PR). If throttled, no new batch is created, and the old
  batch continues — that's the intended behavior. If not throttled, the code proceeds past the throttle to
  CancelSupersededBatches and then creates a new batch. The scenario described (new commit throttled while stale
   batch runs) is working as designed: the throttle deliberately lets the existing review finish rather than
  canceling it for nothing.

  Finding 3 (Medium): "Nullable GitHub PR author can break polling" — Invalid. As discussed in an earlier review
   round, Go's encoding/json does not fail when unmarshaling "author": null into a non-pointer struct. It leaves
   the field at its zero value (ghPRAuthor{Login: ""}). No decode error, no poll failure. This is documented Go
  behavior.

  Finding 4 (Medium): "TOCTOU race in CancelClosedPRBatches" — Technically true but not worth fixing. The race
  window is between the SELECT (line 482) and DELETE (line 527) within the same function call, which in SQLite
  WAL mode is milliseconds. For the race to matter, a batch would need to be claimed for synthesis in that
  narrow window AND the PR would need to be confirmed closed by isPROpen. Even if it fires, the worst outcome is
   a redundant comment post fails (since the batch gets deleted), and the fail-safe UnclaimBatch path handles
  that. Adding a conditional DELETE would add complexity for a near-impossible scenario with benign
  consequences. Not actionable."

PR #411, opened 2026-03-02
Add configurable PR comment upsert and fix UTF-8 truncation

dannysteenman:
## Summary

CI review comments are always created as new comments. This adds an opt-in `ci.upsert_comments` setting that finds the existing roborev marker comment and patches it in-place instead.

- Add `ci.upsert_comments` config (global and per-repo, default false)
- Add `FindExistingComment`, `UpsertPRComment`, and `CreatePRComment` in `internal/github`
- Fall back to creating a new comment when PATCH returns 403/404 (token mismatch)
- Target the newest marker comment to reduce fallback churn
- Fix UTF-8 truncation: replace `utf8.ValidString` full-string scan with boundary-only `TrimPartialRune`
- Propagate caller context through `postCIComment` for Ctrl+C cancellation

### Config

```toml
# Global (~/.roborev/config.toml)
[ci]
upsert_comments = true

# Per-repo (.roborev.toml) — overrides global
[ci]
upsert_comments = false
```

### DIFF
--- a/cmd/roborev/ci.go
+++ b/cmd/roborev/ci.go
@@ -267,7 +267,7 @@ func runCIReview(ctx context.Context, opts ciReviewOpts) error {
 		upsert := resolveCIUpsertComments(
 			repoCfg, globalCfg)
 		if err := postCIComment(
-			ghRepo, prNumber, comment, upsert,
+			ctx, ghRepo, prNumber, comment, upsert,
 		); err != nil {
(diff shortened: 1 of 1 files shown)


wesm:
Looking at this. I actually want to see all the review comments from my CI bot rather than updating them. I'm going to take a pass on this and make the upserting behavior configurable rather than the default

PR #429, opened 2026-03-05
Add multi-repo workspace support

wesm:
## Summary

Adds workspace mode: `roborev list` and `roborev review` now work from non-git parent directories that contain multiple git repos. Running `roborev list` from such a directory filters jobs/repos by path prefix, showing results across all child repositories.

### Workspace detection (`cmd/roborev/list.go`, `cmd/roborev/review.go`)

- `roborev list` detects when the working directory is not a git repo and no `--repo` is specified — switches to workspace mode, sending the current directory as a `repo_prefix` query param
- `roborev review` from a non-git parent directory lists child repos containing `.git` (file or directory) and prints hint paths with full absolute paths
- Auto-detected branch is suppressed in workspace mode (filtering one repo's branch across a workspace is meaningless), but explicit `--branch` flags are always honored via `cmd.Flags().Changed("branch")`

### Composable repo query options (`internal/storage/repos.go`)

Replaced three near-identical functions (`ListReposWithReviewCounts`, `ListReposWithReviewCountsByPrefix`, `ListReposWithReviewCountsByBranch`) with a single function using composable options:

- `WithRepoPathPrefix(prefix)` — filter repos by path prefix
- `WithRepoBranch(branch)` — filter repos by branch (switches from LEFT JOIN to INNER JOIN + HAVING)
- Both options can be combined, enabling prefix+branch filtering that was previously impossible

### SQL ESCAPE character fix (`internal/storage/jobs.go`, `internal/storage/repos.go`)

Changed `escapeLike` from `\` to `!` as the ESCAPE character. The backslash conflicted with Windows path separators stored in `root_path`, causing invalid escape sequences in LIKE clauses.

### Cross-platform path normalization

- `GetOrCreateRepo` and `GetRepoByPath` apply `filepath.ToSlash` at write time so DB rows always store forward-slash paths regardless of platform
- `RenameRepo` and `FindRepo` normalize paths before querying
- Server handlers apply `filepath.ToSlash(filepath.Clean(...))` to prefix query params, normalizing trailing slashes, `..` segments, and backslashes
- Client-side (`list.go`) normalizes the workspace prefix to forward slashes before sending

### Git worktree detection (`cmd/roborev/review.go`)

`findChildGitRepos` now detects `.git` as either a file (worktrees, submodules) or a directory (regular repos). Previously only checked `IsDir()`.

### Path normalization in server handlers (`internal/daemon/server.go`)

`handleListJobs` and `handleListRepos` apply `filepath.Clean` + `filepath.ToSlash` to prefix params, and pass both prefix and branch options when present (previously mutually exclusive via if/else chain).

### DIFF
--- a/internal/daemon/server_jobs_test.go
+++ b/internal/daemon/server_jobs_test.go
@@ -1431,5 +1431,13 @@ func TestHandleListJobsSlashNormalization(t *testing.T) {
 				len(resp.Jobs),
 			)
 		}
+		for _, j := range resp.Jobs {
+			if !strings.HasPrefix(j.RepoPath, ws+"/") {
+				t.Errorf(
+					"Job %d repo_path %q should be under %s",
+					j.ID, j.RepoPath, ws,
+				)
+			}
+		}
 	})
 }
--- a/internal/daemon/server_repos_test.go
+++ b/internal/daemon/server_repos_test.go
@@ -278,9 +278,13 @@ func TestHandleListReposSlashNormalization(t *testing.T) {
 	seedRepoWithJobs(t, db, ws+"/repo-y", 1, "ry")
 	seedRepoWithJobs(t, db, tmpDir+"/other-z", 1, "rz")
 
+	type repoEntry struct {
+		Name     string `json:"name"`
+		RootPath string `json:"root_path"`
+	}
 	type reposResponse struct {
-		Repos      []struct{ Name string } `json:"repos"`
-		TotalCount int                     `json:"total_count"`
+		Repos      []repoEntry `json:"repos"`
+		TotalCount int         `json:"total_count"`
 	}
 
 	t.Run("forward-slash prefix matches stored paths", func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


wesm:
Re: review cf7bd80 (job 7939) — both findings were already addressed in earlier rounds:

1. **Root prefix "/"** — `WithRepoPrefix` and `WithRepoPathPrefix` trim trailing slashes via `strings.TrimRight(prefix, "/")`, so `/` becomes empty string and matches all repos. Covered by `TestRootPrefixMatchesAllRepos`.

2. **Windows backslash paths** — `GetOrCreateRepo` and `GetRepoByPath` apply `filepath.ToSlash` at write time (commit 660ce1da), so DB rows always store forward slashes. Prefixes are also normalized to forward slashes in the server handlers. No mismatch possible.

wesm:
Duplicate of #426 — created in error.

wesm:
created by accident

PR #430, opened 2026-03-05
Add cursor agent support, reject unknown agents, fix ACP alias collision

wesm:
## Summary

- Add cursor agent with `"agent"` CLI binary and `"agent" → "cursor"` alias
- Reject unknown agent names (typos like `--agent claud`) with a clear error listing known agents, instead of silently falling back to a different agent
- Fix ACP alias collision where `acp.name = "agent"` would incorrectly intercept cursor requests via alias resolution — `isConfiguredACPAgentName` now uses exact matching on the raw (pre-alias) name
- Return 400 for unknown agent names and 503 for no-agents-available, using typed `UnknownAgentError` so HTTP handlers can distinguish the two cases
- Add deterministic agent fallback/availability tests for both enqueue and fix endpoints, with proper registry and PATH isolation

### DIFF
--- a/cmd/roborev/refine_test.go
+++ b/cmd/roborev/refine_test.go
@@ -285,8 +285,8 @@ func TestSelectRefineAgentCodexACPConfigAliasUsesACPResolution(t *testing.T) {
 func TestSelectRefineAgentCodexFallbackUsesRequestedReasoning(t *testing.T) {
 	t.Cleanup(testutil.MockExecutableIsolated(t, "codex", 0))
 
-	// Request an unavailable agent, codex should be used as fallback
-	selected, err := selectRefineAgent(nil, "nonexistent-agent", agent.ReasoningThorough, "")
+	// Request a known-but-unavailable agent, codex should be used as fallback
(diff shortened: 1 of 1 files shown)


wesm:
"Invalid. The file compiles cleanly and the test passes. The reviewer's agents are mangling Go escape sequences   
  when rendering the source — they interpret \" and \n in the Go string literals as actual quotes and newlines,
  then report the "unescaped" result as a syntax error. The actual strings on disk are correct:                    
                                                                                                                 
  - Line 919: fmt.Sprintf("@\"%s\" %%*\r\n", gitPath) — valid Go                                                   
  - Line 924: fmt.Sprintf("#!/bin/sh\nexec '%s' \"$@\"\n", gitPath) — valid Go                                   
  - Line 978: content = "@exit /b 0\r\n" — valid batch command
"

Not sure what is up with github actions, I'll see if it comes back on later

wesm:
"No, same false positive as before. The review agents are interpreting Go escape sequences (\", \n, $@) instead of
   reading the raw source. The file compiles and passes on all platforms — I verified this earlier and the full go 
  test ./... suite passed."

PR #434, opened 2026-03-06
Fix Claude agent error capture for failed reviews

### LINKED ISSUE #432: Harden error capturing and reporting when claude agent fails to launch
Users have reported seeing "No review output generated" when claude simply failed to do anything (i.e. do to a missing API key or SSO credentials). We need to make sure these errors are surfaced in the review screen so that users know what error was generated rather than having to dig through the log files to find out what's wrong

wesm:
## Summary
- Fix three bugs in Claude's stream-json parser that caused errors (auth failures, API errors, quota exhaustion) to be silently swallowed, producing empty "Fail" reviews instead of actionable error messages
- Handle `is_error` flag on result events, parse content block arrays (real Claude Code format), and return placeholder text for empty output consistent with other agents
- Reject placeholder output in compact validation so compact jobs don't close source reviews with useless content
- Add matching Cursor agent coverage since it shares `parseStreamJSON`

Closes #432

## Test plan
- [x] `TestParseStreamJSON` — 10 new subcases: `is_error` result events (with error message, with result text, no details, with partial assistant output), content block arrays (single, multiple, mixed with tool_use), standalone error events, `is_error=false` normal path
- [x] `TestExtractContentText` — 8 cases for string vs array content extraction
- [x] `TestClaudeReviewEmptyOutput` — end-to-end: valid events, no content → returns "No review output generated"
- [x] `TestClaudeReviewErrorResult` — end-to-end: `is_error` result with exit 0 → error propagated
- [x] `TestClaudeReviewErrorResultNonZeroExit` — end-to-end: `is_error` + non-zero exit → includes stream error and exit status
- [x] `TestCursorReviewEmptyOutput`, `TestCursorReviewErrorResult`, `TestCursorReviewErrorResultNonZeroExit` — mirror Claude cases for Cursor
- [x] `TestIsValidCompactOutput` — rejects placeholder text, empty, error prefixes; accepts real reviews
- [x] Full test suite passes (`go test ./...`)

### DIFF
--- a/internal/agent/cursor_test.go
+++ b/internal/agent/cursor_test.go
@@ -181,6 +181,68 @@ func TestCursorReviewPipesPromptViaStdin(t *testing.T) {
 	assertFileNotContains(t, mock.ArgsFile, prompt)
 }
 
+func TestCursorReviewEmptyOutput(t *testing.T) {
+	a, _ := setupMockCursorAgent(t, MockCLIOpts{
+		StdoutLines: []string{`{"type":"system","subtype":"init"}`},
+	})
+
+	result, err := a.Review(
+		context.Background(), t.TempDir(), "abc123", "review this", nil,
+	)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if result != "No review output generated" {
+		t.Errorf("expected %q, got %q", "No review output generated", result)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


wesm:
"This is a false positive. Line 56 (output = strings.TrimSpace(output)) already normalizes the input before the
   comparison on line 63. Any trailing whitespace or newlines are stripped before the exact match runs. No
  change needed."

PR #440, opened 2026-03-06
feat: excluded commit patterns

### LINKED ISSUE #423: Support skipping commits based on message content
It would be awesome if we could configure `excluded_commits` or similar, to skip reviewing commits that include things like `[skip ci]` (similar to [how GitHub Actions does](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/)). Being able to provide substrings that appear anywhere in the commit message might be helpful, but even requiring full matching like `excluded_branches` would work.

roryschadler:
## Summary

  - Add `excluded_commit_patterns` config option that allows skipping reviews for commits whose message contains any of the configured substrings (e.g., `[skip review]`,
  `[wip]`)
  - Matching is case-insensitive and checks the full commit message (subject + body)
  - Returns `{"skipped": true, ...}` when a commit matches an excluded pattern

Fixes #423

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -537,18 +537,19 @@ type RepoCIConfig struct {
 
 // RepoConfig holds per-repo overrides
 type RepoConfig struct {
-	Agent              string   `toml:"agent"`
-	Model              string   `toml:"model"` // Model for agents (format varies by agent)
-	BackupAgent        string   `toml:"backup_agent"`
-	BackupModel        string   `toml:"backup_model"`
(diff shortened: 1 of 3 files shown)


roryschadler:
^ I think this behavior might be intentional - if not, I'm happy to resolve. An initial pass makes me think that change would extend the scope of this PR pretty significantly, moving diff size-checking and SHA resolution around. Again, if that's desirable, I'm happy to make that change!

wesm:
superseded by https://github.com/roborev-dev/roborev/pull/454

PR #449, opened 2026-03-06
Fix agent resolution and improve skill definitions

wesm:
## Summary

- Add variadic `backups ...string` parameter to `GetAvailable` and `GetAvailableWithConfig` so configured backup agents are tried before the hardcoded codex→claude-code→gemini→... fallback chain
- Update all call sites (handleEnqueue, handleFixJob, CI poller, review, fix, analyze, refine, batch) to resolve the backup agent from config via `ResolveBackupAgentForWorkflow` and pass it through
- When the backup agent is selected and no explicit model was requested, use `ResolveBackupModelForWorkflow` for the model instead of the default model
- Fix backup agent in refine and review workflows inheriting the primary agent's model instead of resolving its own backup model
- Extract shared `applyModelForAgent()` helper used by both refine and review for backup-vs-primary model resolution, replacing inline duplicated logic
- Add regression tests that call `applyModelForAgent` directly to verify backup agent model inheritance and empty-model guard
- Add invocation guards to review skills preventing false triggers when users paste existing review results
- Make fix skill check conversation context before running discovery commands, skipping fetch when findings are already present; require matching pasted findings to jobs by commit SHA before closing
- Replace "ask to commit" in fix skill with deference to project CLAUDE.md conventions
- Add general principle to all skills: treat instructions as guidelines, use conversation context, defer to project-level config on conflicts

### DIFF
--- a/cmd/roborev/refine_test.go
+++ b/cmd/roborev/refine_test.go
@@ -1050,6 +1050,13 @@ func TestApplyModelForAgent_SameAgentPrimaryAndBackup(t *testing.T) {
 		t.Fatalf("agent.Get: %v", err)
 	}
 
+	// Config with distinct primary and backup models so we can tell
+	// which resolution path was taken.
+	cfg := &config.Config{
+		ReviewModel:       "primary-model",
+		ReviewBackupModel: "backup-model",
+	}
+
 	// Primary and backup are the same agent. The helper should NOT
 	// treat this as "using backup" — the CanonicalName comparison
 	// against preferredAgent must prevent that.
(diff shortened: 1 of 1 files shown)


wesm:
"Not valid. Every agent's WithModel is a no-op when the model is empty:

  func (a *CodexAgent) WithModel(model string) Agent {
      if model == "" {
          return a  // <-- preserves existing model
      }
      ...
  }

  The worker calls .WithModel(job.Model) at worker.go:403, but when job.Model is "", the agent is returned unchanged — its preconfigured
  default model is preserved. This is already tested explicitly in TestWithModelEmptyPreservesDefault (agent_test.go:272).

  The finding's premise — that WithModel("") "wipes an agent's preconfigured model" — is wrong. False positive.
"

PR #457, opened 2026-03-07
Fix daemon startup restart loop

wesm:
## Summary

- Fix a race where CLI clients discover the daemon via runtime files before the HTTP server is accepting connections, causing repeated unnecessary restarts
- Bind the TCP listener and wait for server readiness before publishing runtime metadata, eliminating the window where clients see a "zombie" daemon
- Add a lightweight `/api/ping` endpoint for reliable liveness probes, with fallback to `/api/status` for older daemons
- Refactor `ensureDaemon()` to use the new `ProbeDaemon` helper instead of raw HTTP calls, and prefer the version field from runtime files when available
- Validate that the daemon bind address is loopback before starting, rejecting wildcard addresses like `0.0.0.0` or `:7373`
- Move hook-upgrade warnings to a background goroutine so they don't block daemon startup

### DIFF
--- a/cmd/roborev/daemon_lifecycle.go
+++ b/cmd/roborev/daemon_lifecycle.go
@@ -132,12 +132,15 @@ func ensureDaemon() error {
 
 	// First check runtime files for any running daemon
 	if info, err := getAnyRunningDaemon(); err == nil {
-		daemonVersion := info.Version
-
 		if !skipVersionCheck {
-			if probe, err := daemon.ProbeDaemon(info.Addr, 2*time.Second); err == nil {
-				daemonVersion = probe.Version
+			probe, err := daemon.ProbeDaemon(info.Addr, 2*time.Second)
+			if err != nil {
+				if verbose {
+					fmt.Printf("Daemon probe failed, restarting...\n")
+				}
+				return restartDaemonForEnsure()
 			}
+			daemonVersion := probe.Version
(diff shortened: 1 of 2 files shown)


wesm:
" - High: partially valid as an API-edge, but not as the concrete daemon run regression it claims. In internal/daemon/server.go:192, Start() will wait on
    serveErrCh if startup is canceled before readiness and the serve goroutine has not exited yet. That can hang if some caller cancels ctx and never calls
    Stop(). But the real daemon entrypoint does call server.Stop() after cancellation in a shutdown goroutine at cmd/roborev/daemon_cmd.go:197, specifically so
    Start() does not return before shutdown cleanup and FinalPush() complete. The suggested fix would break that synchronization.
  - Medium: invalid. The runtime-file branch in cmd/roborev/daemon_lifecycle.go:133 uses getAnyRunningDaemon, which is daemon.GetAnyRunningDaemon. That function
    already requires the daemon to be responsive via IsDaemonAlive() and ProbeDaemon() in internal/daemon/runtime.go:204. So ROBOREV_SKIP_VERSION_CHECK=1 skips
    version enforcement, not liveness validation.

  If you want the High issue turned into an actionable item, I’d restate it as: Server.Start(ctx) has an unclear contract for callers that cancel ctx without
  also driving Stop()."

PR #461, opened 2026-03-07
tighten review prompts, simplify verdict parsing, and isolate CI poller tests

wesm:
## Summary
- make CI poller tests hermetic by defaulting synthesis to the in-process `test` agent in the test harness
- simplify verdict parsing to use only deterministic signals: structured severity/findings markers fail, clear pass phrases pass, and ambiguous prose defaults to fail
- document in code and tests that we are not trying to parse narrative caveats after `No issues found.` and that output-shaping problems should be fixed in review prompts instead
- add Codex and Claude Code review templates that require final-results-only output, forbid process narration/front matter, and make the review format stricter
- update prompt tests and verdict tests to lock in the new prompt and parser policy

## Why
The previous direction was drifting toward sentence-level natural-language parsing of review prose. That made verdict handling brittle and encouraged compensating in the parser for output-shape problems that are better solved at the prompt layer. This change tightens both sides: prompts now ask for cleaner review output, and verdict parsing stays intentionally simple and deterministic.

## Testing
- go fmt ./...
- go vet ./...
- go test ./internal/daemon -run TestCIPollerPostBatchResults_SetsFailureStatusOnMixedOutcome -count=5 -timeout=30s
- go test ./internal/daemon -count=1 -timeout=2m
- go test ./internal/prompt
- go test ./... -count=1

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -51,4 +51,17 @@ default branch has no config (no remote, config not yet committed). The
 daemon never checks out PR branches — the working tree is always the
 operator's own checkout. Do not flag filesystem fallback as a trust-boundary
 issue or claim that PR authors can influence config via the working tree.
+
+## Verdict parsing and review output shape
+
+Verdict parsing is intentionally simple and deterministic. Prefer explicit
+structured signals such as severity labels, findings sections, and clear pass
+phrases. Do not ask for elaborate natural-language contradiction parsing of
+free-form prose after "No issues found." or similar pass text.
+
+If agent review output is too chatty, includes process narration, or mixes
+findings with summary/front matter in a way that makes verdict detection less
+reliable, the fix should be to tighten the review prompts/templates and output
+contract. Do not ask for increasingly broad deterministic heuristics to parse
+arbitrary narrative text.
 """


wesm:
From Codex

"Yes, I agree we should leave it as is.

  That review is arguing for reopening the exact tradeoff we just intentionally closed. The whole
  point of the simplification was:

  - keep verdict parsing deterministic and small
  - stop trying to interpret contradictory natural-language prose
  - push output-shape discipline into prompts instead of growing parser heuristics

  The suggested remediation is basically “reintroduce caveat parsing, but a little narrower this
  time.” That will drift back into the same brittle space.

  The only parser changes I’d support from here are stronger structured signals, for example:

  - explicit Verdict: Pass|Fail
  - stricter Review Findings: section conventions
  - eventually a machine-readable format for reviews

  But for free-form prose like No issues found, but ..., I would not add another contradiction
  detector. If the agents still emit that shape after the prompt tightening, that’s a prompt/
  template quality issue, not a verdict-parser issue.

  If you want, I can add a PR comment saying we intentionally chose not to parse prose caveats and
  that follow-up work should focus on stricter structured review output.
"

wesm:
These changes are as intended. Merging

PR #474, opened 2026-03-08
Hide daemon status from review view

wesm:
## Summary
- remove the daemon status line from the review detail header
- keep the queue view daemon status behavior unchanged
- update review rendering tests and visible-line expectations for the restored layout

## Testing
- go fmt ./...
- go vet ./...
- go test ./cmd/roborev/tui
- go test ./...


wesm:
The review findings are irrelevant. The Daemon version line was added mistakenly to the review screen earlier today

PR #476, opened 2026-03-08
Harden job logging and timeout cleanup

wesm:
## Summary

This PR hardens two failure paths that show up under resource pressure:

1. Per-job log files could be missing even though the job appeared in the TUI.
2. Jobs could appear to run well past `job_timeout_minutes` because the worker context timed out but the agent subprocess/pipe handling did not always unwind promptly.

The primary goal here is reliability, not changing policy. The existing default 30 minute timeout remains in place; this change makes that timeout and per-job logging more dependable.

### Scenario 1: log file creation fails when the job starts

Per-job logging was best-effort. We attempted to open `~/.roborev/logs/jobs/<id>.log` once at job start. If that failed because of a transient filesystem or resource problem, roborev logged a warning and permanently disabled disk logging for that job.

On a resource-constrained machine, that can happen if:
- the machine is under RAM pressure and filesystem operations stall or fail transiently
- the data directory is temporarily unavailable or in a bad state
- writes start failing mid-job and never recover into a reopened log file

The visible result is a job in the TUI with no corresponding job log file, even though the job itself may finish.

### Scenario 2: the job timeout fires, but the review does not actually unwind

The worker already applies a hard timeout via `context.WithTimeout(...)`, defaulting to 30 minutes. The bug is that some agent subprocess paths could still remain stuck in process wait / pipe handling after the context deadline.

That means a user can observe a review job apparently "running" for longer than the configured timeout, especially when subprocesses are slow to terminate or inherited pipes remain open.

On low-resource machines this is more likely because:
- child processes can stall while under memory pressure
- cleanup can take longer
- stdout pipe / background process behavior can keep `Wait()` from returning promptly

### 1. Retryable job log writer

Replace the old one-shot `safeWriter` behavior with a retrying `jobLogWriter`:
- it keeps trying to open/reopen the per-job log file after transient failures
- it buffers a bounded amount of output in memory while disk logging is unavailable
- it flushes buffered content once logging becomes available again
- if the buffer overflows, it records how many bytes were dropped instead of silently disabling logging forever

This means a transient log-file open/write failure no longer dooms the entire job to have no disk log.

### 2. Centralized subprocess wait configuration and timeout cleanup

Add shared subprocess helpers so all agent adapters use the same wait-delay handling.

For streaming adapters, close the stdout pipe when the job context is done. This helps break cases where the context deadline has fired but the reader / wait path is still blocked behind lingering pipe state.

After `cmd.Wait()` completes, return `context.DeadlineExceeded` when the job context has expired instead of surfacing a generic subprocess error.

### 3. Explicit timeout handling in the worker

When an agent returns a deadline error, the worker now records a stable job error like:

`agent timeout after 30m0s`

and sends it through the normal retry/failover path.

This makes timeout failures clearer in the database, TUI, and hooks, and avoids jobs looking like ambiguous generic agent failures.

## Why this fixes the reported behavior

The reported symptoms were:
- not every displayed job had a log file
- some review jobs seemed to hang for over an hour
- the machine had limited RAM

This PR directly addresses those failure modes:
- transient resource issues no longer permanently disable per-job disk logging
- timed-out agent jobs are much more likely to unwind promptly instead of lingering past the configured timeout

## Validation

- `go fmt ./...`
- `go vet ./...`
- `go test ./...`

## Notes from local investigation

On the investigated machine there are more jobs in `reviews.db` than files in `~/.roborev/logs/jobs`, but the mismatch is historical rather than current: jobs from the last 7 days all had logs. The older missing logs line up with the existing log-retention/cleanup behavior and prior best-effort logging behavior. This PR is aimed at preventing new gaps under transient resource pressure and making timeout enforcement behave consistently.

### DIFF
--- a/internal/agent/process.go
+++ b/internal/agent/process.go
@@ -24,8 +24,11 @@ func configureSubprocess(cmd *exec.Cmd) *subprocessTracker {
 	if cmd.Cancel != nil {
 		cancel := cmd.Cancel
 		cmd.Cancel = func() error {
-			tracker.canceledByContext.Store(true)
-			return cancel()
+			err := cancel()
+			if err == nil {
+				tracker.canceledByContext.Store(true)
+			}
+			return err
 		}
 	}
 	return tracker
--- a/internal/agent/process_test.go
+++ b/internal/agent/process_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io/fs"
+	"os"
 	"os/exec"
 	"sync/atomic"
 	"testing"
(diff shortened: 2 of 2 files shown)


wesm:
"exec.CommandContext does not leave `cmd.Cancel` nil. In Go, `exec.CommandContext` sets
  `cmd.Cancel` to kill the process by default. All subprocesses in this change use
  `exec.CommandContext`, so the wrapper in `configureSubprocess` is active in the real code paths. I
  verified this locally with both `go doc os/exec.CommandContext` and a small runtime check showing
  `cmd.Cancel == nil` is false for `exec.CommandContext(...)`. Because of that, the reported premise
  does not apply to this branch.`"

PR #483, opened 2026-03-08
Document auto-written TOML config files

mariusvniekerk:
## Summary
- switch global config writes to `go-toml/v2` so auto-written TOML can include inline comments
- route `roborev config set` global and local writes through typed save helpers so both `config.toml` and `.roborev.toml` get documented output
- update `roborev init --agent ...` to create a commented repo config scaffold and add regression coverage for the global, local, and init paths

## Test Plan
- go vet ./...
- go test ./...


wesm:
invalid. merging

PR #488, opened 2026-03-09
Add TUI control socket, --no-quit flag, and runtime metadata

wesm:
## Summary

- Add a newline-delimited JSON protocol over a Unix domain socket (`tui.{PID}.sock`) that allows external tools to programmatically query and mutate running TUI instances
- Support 13 commands: 4 queries (`get-state`, `get-filter`, `get-jobs`, `get-selected`) and 9 mutations (`set-filter`, `clear-filter`, `set-hide-closed`, `select-job`, `set-view`, `close-review`, `cancel-job`, `rerun-job`, `quit`)
- Write runtime metadata to `tui.{PID}.json` (PID, socket path, server address) for discoverability; refresh on daemon reconnect when server address changes
- Add `--control-socket` flag for custom socket paths and `--no-quit` flag that suppresses keyboard quit (`q`) in queue/tasks views for managed TUI instances
- `quit` control command terminates the TUI regardless of `--no-quit`, allowing external controllers to manage lifecycle programmatically
- `set-filter` resolves display names (e.g. "msgvault") to root paths via `/api/repos`, matching the interactive filter modal's resolution; `reposMsg` carries an explicit `branchFiltered` flag so branch-scoped loads don't clobber the display-name map
- Help text conditionally omits `q`/quit references when `--no-quit` is active
- Socket security: stale sockets from dead processes cleaned via platform-specific PID checks (signal-0 on Unix, tasklist on Windows); socket created with `0600` permissions via `os.Chmod` (not process-wide umask); managed socket directory tightened to `0700`; custom `--control-socket` paths skip directory chmod to avoid clobbering shared directories like `/tmp`
- Startup safety: control listener deferred until Bubble Tea event loop is running (`ready` channel); `controlSocketReadyMsg` prevents advertising a socket path before the listener binds; `cleanupDone` channel ensures socket removal completes before `Run()` returns; `programDone` channel prevents deadlock if program exits before first `Update`
- Cleanup ordering: socket file removed before listener close to prevent a successor process's socket from being unlinked by a dying predecessor
- Filter mutations (`set-filter`, `clear-filter`) validate all lock constraints before mutating any state to prevent partial application
- `rerun-job` clears `Closed`/`Verdict` fields so rerun jobs remain visible under `hideClosed`; both the control socket and keyboard rerun paths use a shared `rerunSnapshot` for rollback on failure
- `select-job` rejects jobs hidden by active filters; `close-review`/`cancel-job` only reflow selection when the mutated job is the currently selected row

### DIFF
--- a/cmd/roborev/tui/control_test.go
+++ b/cmd/roborev/tui/control_test.go
@@ -693,6 +693,52 @@ func TestHandleCtrlCancelJob_WrongStatus(t *testing.T) {
 	require.False(t, resp.OK, "expected error for non-cancellable job")
 }
 
+func TestHandleCancelKey_ClearsSelectionWhenNoneVisible(t *testing.T) {
+	m := newModel(testServerAddr, withExternalIODisabled())
+	m.currentView = viewQueue
+	m.hideClosed = true
+	m.jobs = []storage.ReviewJob{
+		makeJob(1, withStatus(storage.JobStatusRunning)),
+	}
+	m.selectedIdx = 0
+	m.selectedJobID = 1
+
+	result, cmd := m.handleCancelKey()
+	updated := result.(model)
+	assert.Equal(t, storage.JobStatusCanceled, updated.jobs[0].Status)
+	assert.Equal(t, -1, updated.selectedIdx,
+		"selectedIdx should be cleared when no visible jobs remain")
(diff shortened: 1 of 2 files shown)


wesm:
Acceptable risk. Merging

PR #495, opened 2026-03-10
Honor *_cmd config values for agent availability checks

### LINKED ISSUE #381: claude_code_cmd config value is ignored — agent always invokes hardcoded "claude"
## Description

The `claude_code_cmd` field in `~/.roborev/config.toml` is parsed into the config struct but never actually used when invoking the Claude Code agent. The agent always runs the hardcoded binary name `"claude"` regardless of what is configured. The same issue affects `codex_cmd` and `cursor_cmd`.

## Steps to Reproduce

1. Set `claude_code_cmd = "/path/to/custom-claude-wrapper"` in `~/.roborev/config.toml`
2. Run `roborev daemon restart`
3. Trigger a review
4. Observe via `ps` or process inspection that `claude` is invoked, not the configured command

## Root Cause

In `internal/agent/claude.go`, the agent is registered at package init time with an empty command:

```go
// claude.go:301-303
func init() {
    Register(NewClaudeAgent(""))  // always uses hardcoded "claude"
}
```

`NewClaudeAgent("")` falls back to the hardcoded default:

```go
// claude.go:31-35
func NewClaudeAgent(command string) *ClaudeAgent {
    if command == "" {
        command = "claude"
    }
    ...
}
```

Meanwhile, `config.go` defines and parses `ClaudeCodeCmd` (default `"claude"`) but the value is never read and never passed to `NewClaudeAgent()`.

## Expected Behavior

When `claude_code_cmd = "/path/to/custom-wrapper"` is set, Roborev should invoke that command instead of `claude`.

## Suggested Fix

Pass the config value when registering the agent at daemon startup (rather than in `init()`), e.g.:

```go
Register(NewClaudeAgent(cfg.ClaudeCodeCmd))
```

## Motivation

This feature is needed to wrap the Claude invocation with per-caller environment variables — for example, to use different audio notification packs via [Peon-Ping](https://github.com/PeonPing/peon-ping)'s `CLAUDE_PEON_DIR` depending on whether Claude is being run interactively or triggered by Roborev. Without a working `claude_code_cmd`, there is no supported way to distinguish a Roborev-triggered Claude session from an interactive one at the hook level.

wesm:
## Summary

- Extract the inline command override switch in `GetAvailableWithConfig` into a reusable `applyCommandOverrides` helper that clones an agent and sets the configured command path
- Add `isAvailableWithConfig` that checks agent availability using the configured command before falling back to the hardcoded default, so agents are found even when only the configured binary is in PATH
- `GetAvailableWithConfig` now checks the preferred agent and backup agents with config-aware availability before falling through to the `GetAvailable` fallback chain
- Add tests for `claude_code_cmd`, `codex_cmd`, `cursor_cmd`, and `pi_cmd` overrides, including the case where only the configured command is available

Fixes #381

### DIFF
--- a/internal/agent/acp.go
+++ b/internal/agent/acp.go
@@ -1591,7 +1591,22 @@ func GetAvailableWithConfig(preferred string, cfg *config.Config, backups ...str
 			}
 		}
 
-		// Finally fall back to normal auto-selection (with backups).
+		// ACP unavailable — try backup agents with config-aware
+		// availability so *_cmd overrides are honored.
+		if cfg != nil {
+			for _, b := range backups {
+				b = resolveAlias(b)
+				if b == "" {
+					continue
+				}
+				if _, ok := registry[b]; ok && isAvailableWithConfig(b, cfg) {
+					a, _ := Get(b)
+					return applyCommandOverrides(a, cfg), nil
+				}
+			}
+		}
+
+		// Finally fall back to normal auto-selection.
 		return GetAvailable("", backups...)
 	}
 
--- a/internal/agent/acp_test.go
+++ b/internal/agent/acp_test.go
@@ -1149,6 +1149,49 @@ func TestGetAvailableWithConfigPiCmd(t *testing.T) {
 	assert.Equal(t, filepath.Join(fakeBin, wrapper), ca.CommandName())
 }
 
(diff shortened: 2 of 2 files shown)


wesm:
False positives. Merging

PR #497, opened 2026-03-10
Fix post-commit hook to use main repo root in git worktrees

mariusvniekerk:
## Summary

- When the post-commit hook fires from a linked git worktree, `GetRepoRoot` returns the worktree path rather than the main repo root, causing the daemon to register the commit under a phantom repo entry
- Fix by using `GetMainRepoRoot` for the `RepoPath` sent to the daemon, while keeping `GetRepoRoot` (worktree-local path) for rebase detection and branch/ref resolution, since those states live in the worktree-specific git dir
- Adds table-driven tests covering both plain repo and worktree cases for `RepoPath` correctness and rebase suppression during in-progress rebases

### DIFF
--- a/cmd/roborev/postcommit_test.go
+++ b/cmd/roborev/postcommit_test.go
@@ -295,16 +295,20 @@ func TestPostCommitSendsLocalRepoPath(t *testing.T) {
 // end-to-end shell hook flow.
 func TestPostCommitSkipsEnqueueDuringRebase(t *testing.T) {
 	tests := []struct {
-		name string
-		dir  string // directory to create inside .git/
+		name     string
+		setup    func(t *testing.T) repoUnderTest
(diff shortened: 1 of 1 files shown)


wesm:
" Not valid as a correctness issue. Since Go 1.22, exec.Cmd.Env has guaranteed last-wins semantics for
  duplicate keys — Go deduplicates before passing to the OS. The append(os.Environ(), ...) pattern
  works because the overrides are appended at the end and are guaranteed to win. The project uses Go
  1.24.

  The TestGitRepo.Run dedup was written before this guarantee existed (or for extra clarity), but the
  behavior is identical either way. Changing the test to match that style would be a consistency
  preference, not a bug fix. I'd skip it to avoid churn on an already large PR.
"

PR #500, opened 2026-03-11
Fix post-commit hook not firing in linked worktrees

wesm:
## Summary

- Resolve relative `core.hooksPath` values against the main repository root instead of the worktree root, both in `GetHooksPath` (for roborev's internal use) and via `EnsureAbsoluteHooksPath` (to fix the stored git config so git itself dispatches hooks correctly from linked worktrees)
- Add JSONL post-commit hook logging to `~/.roborev/post-commit.log` so that silent hook failures leave an audit trail (timestamp, repo, outcome, reason)
- Preserve git tilde-expansion paths (`~`, `~/path`, `~user/path`) during normalization
- Apply hooks-path normalization in `init`, `install-hook`, and `uninstall-hook` as a fatal error so these commands stop before operating on the wrong directory
- Clean up test anti-patterns in `internal/git/git_test.go`

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -234,7 +234,7 @@ func GetMainRepoRoot(path string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("git rev-parse --git-dir: %w", err)
 	}
-	gitDir := strings.TrimSpace(string(gitDirOut))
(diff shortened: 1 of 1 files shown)


wesm:
Enough. merging

PR #501, opened 2026-03-12
Fix roborev fix discovering unrelated reviews in worktrees

wesm:
## Summary

- After fetching candidate jobs from the daemon API, `roborev fix --open`, `--list`, and `--batch` now filter jobs to only those relevant to the current worktree
- Single-commit SHA refs are checked via `git merge-base --is-ancestor` against the worktree HEAD
- Range refs parse the end commit and check its reachability
- Dirty/empty refs compare the job's stored branch against the current worktree branch
- All paths fail open on git errors or missing metadata to avoid silently dropping work
- Adds `TestFilterReachableJobs` (14 subtests) and `TestRunFixOpenFiltersUnreachableJobs` integration test

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -82,6 +82,15 @@ Examples:
 `,
 		Args: cobra.ArbitraryArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


wesm:
False positive

PR #505, opened 2026-03-12
Exclude common lockfiles from reviews, add configurable exclude patterns

### LINKED ISSUE #469: Exclude common lockfiles from reviews, and add repo-level ignore patterns to .roborev.toml
Repos need the equivalent of .gitignore for their reviews, and common lockfiles (package-lock.json, cargo.lock, etc.) should typically be excluded from reviews (there is a pedantic concern about lockfiles introducing security issues, but that should be addressed separately).

related to #467

wesm:
## Summary

- Exclude common lockfiles and generated files from review diffs by default (package-lock.json, yarn.lock, uv.lock, Cargo.lock, go.sum, etc.) using git pathspec exclusions
- Add configurable `exclude_patterns` in both global config (`~/.roborev/config.toml`) and per-repo config (`.roborev.toml`) supporting filenames, directory names, and glob patterns including `**`
- Repo-level exclude patterns are read from the default branch (not the working tree) to prevent untrusted branches from suppressing files — except for dirty reviews which use working-tree config via `ResolveExcludePatternsLocal`
- Security reviews skip repo-level exclude patterns entirely so a compromised `.roborev.toml` cannot hide files from security review
- Address/refine prompts reconstruct the original diff without user exclude patterns so findings remain valid even if excludes change after the review
- `formatExcludeArgs` emits both `**/name` and `**/name/**` pathspecs for plain and slashed patterns so both files and directory subtrees are excluded consistently
- All exclude filtering (tracked and untracked files) uses git pathspecs as the single source of truth — the in-memory matching functions were removed after unifying on `git ls-files` with pathspec excludes for dirty diffs
- Per-job prompt builders (`NewBuilderWithConfig`) replace the shared builder to avoid mutable state across concurrent workers

Closes #469

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -569,42 +569,38 @@ var excludedPathPatterns = []string{
 	":(exclude,glob)**/.cache/**",
 }
 
-// normalizePattern strips whitespace and leading/trailing slashes
-// from a user-provided exclude pattern. Returns "" for empty input.
-func normalizePattern(p string) string {
-	p = strings.TrimSpace(p)
-	p = strings.TrimRight(p, "/")
-	p = strings.TrimLeft(p, "/")
-	return p
-}
-
 // formatExcludeArgs converts user-provided exclude patterns (filenames
 // or globs) into git pathspec arguments. Plain names without path
 // separators get both **/name (file match) and **/name/** (directory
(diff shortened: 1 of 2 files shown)


wesm:
I think this is good enough here

PR #534, opened 2026-03-18
Fix --all-branches and --branch filtering from worktrees

wesm:
## Summary

- Fix `roborev fix --all-branches` and `--branch <name>` not discovering jobs from other branches when run from a git worktree — `filterReachableJobs` was called with an empty branch override, causing it to filter by the worktree's branch instead of the requested branch
- Deprecate `--open` and `--unaddressed` flags — open job discovery is now the default behavior when no positional job IDs are provided; both flags are hidden and silently ignored for backwards compatibility
- Thread `allBranches` and `explicitBranch` through `runFixOpen`/`runFixBatch` to distinguish three filtering modes: default (commit-graph reachability), `--branch X` (branch-field matching for cross-branch fixing), and `--all-branches` (skip filtering)
- Add tests for all-branches discovery, explicit branch filtering, and worktree reachability

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -159,13 +159,16 @@ Examples:
 						}
 						effectiveBranch = git.GetCurrentBranch(repoRoot)
 					}
-					return runFixBatch(cmd, nil, effectiveBranch, newestFirst, opts)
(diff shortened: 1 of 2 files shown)


wesm:
Non-issues in practice. Merging

PR #537, opened 2026-03-18
Speed up test suite across 4 packages (~15s → ~6s)

cpcloud:
## Summary

- **Isolate agent, config, and review tests from global git config** — extends #531 to three more packages. Adds `TestMain` calling `testenv.RunIsolatedMain()` so `git` operations in tests never trigger the host's `gpg-agent`/`pinentry` (which can add 1-2s per invocation on machines with GPG signing configured).

- **Make worker pool idle sleeps responsive to shutdown** — replaces `time.Sleep` in the worker idle loop with a `select` on a shutdown channel, so tests that stop the daemon don't block waiting for the sleep to finish.

- **Stub `isPROpenFn` in CI poller test harness** — avoids shelling out to `gh` CLI during tests. Removes a source of ~0.5s latency and flaky failures when `gh` isn't authenticated.

- **Parallelize daemon subtests** — marks 21 independent daemon tests as `t.Parallel()`, including `TestIsDaemonAliveLegacyStatusCodes` subtests and the bulk of the worker/failover/cancellation tests. Drops `internal/daemon` from ~7.2s to ~4.2s.

- **Tighten `requireNever` timeout** — the invalid-config watcher test was waiting 500ms for a "never" assertion; 100ms is sufficient and shaves ~400ms.

- **Fix `configureSubprocess` to always set `Cancel` on `exec.Cmd`** — when called on a plain `exec.Command` (no context), `cmd.Cancel` is nil, so the wrapper's `cancel()` call panicked. Now sets a default `Kill` cancel before wrapping. This is a correctness fix surfaced by the timing changes below.

- **Speed up agent subprocess tests (8s → 1s)** — two changes:
  1. Override `subprocessWaitDelay` from 5s to 50ms in tests that don't need the production grace period.
  2. Add a Go 1.25 ETXTBSY probe guard (`case "$1" in *etxtbsy*) exit 0;; esac`) to shell-script test fixtures.

### Go 1.25 ETXTBSY probe — why the guard is necessary

Go 1.25 added a built-in [ETXTBSY probe](https://go.dev/cl/528438) to `os/exec`: every `Cmd.Start()` call forks a child that runs `execve(path, [path, "--help-probe-etxtbsy"])`. If `execve` succeeds (file isn't busy), the probe child **actually executes the script** — it isn't killed, and the parent waits for it to exit before proceeding.

For shell scripts, this means the script runs **twice**: once as the probe (with `$1 = "--help-probe-etxtbsy"`), and once as the real invocation. If the script ignores `$1` and does something slow (like `sleep 5`), the probe child runs the full body, and Go waits for it — adding the script's entire runtime as invisible overhead to `Cmd.Start()`.

This was confirmed via `strace -f`: two child PIDs ran the same `clock_nanosleep(5s)`, one from the probe fork (`CLONE_VFORK|CLONE_PIDFD`) and one from the real execution. The probe child was waited on via `waitid(P_PIDFD)`.

The guard pattern — `case "$1" in *etxtbsy*) exit 0;; esac` as the first line after the shebang — makes the probe child exit immediately. This is only needed in test fixtures; real agent binaries already handle unknown flags gracefully.

Further details are documented in `internal/agent/agent_test_helpers.go` on the `writeTempCommand` function.

### DIFF
--- a/internal/agent/agent_test_helpers.go
+++ b/internal/agent/agent_test_helpers.go
@@ -54,6 +54,39 @@ func skipIfWindows(t *testing.T) {
 	}
 }
 
+// writeTempCommand writes script to a temporary executable file and returns
+// its path. The file is cleaned up when the test finishes.
+//
+// # ETXTBSY probes and Go 1.25
+//
+// Go 1.25 added a built-in ETXTBSY probe to os/exec: every exec.Cmd.Start()
+// forks a child that calls execve(path, ["path", "--help-probe-etxtbsy"]) to
+// check whether the file is still being written. If execve succeeds (no
+// ETXTBSY), the probe child has *actually started running the command* — it
+// is not killed or cleaned up, and the parent waits for it to exit.
+//
+// For most commands this is harmless because the probe exits quickly. But
+// shell scripts that ignore their arguments will execute their full body
+// inside the probe child. A script containing "sleep 60" will spawn a probe
+// process that sleeps for 60 real seconds, and the Go test binary blocks
+// until it finishes (confirmed via strace: the probe child runs
+// clock_nanosleep for the full duration, and the test binary's waitid blocks
+// on it).
+//
+// This means every shell script launched via os/exec in Go 1.25 runs twice:
+// once as the ETXTBSY probe, once for real. Additionally, writeTempCommand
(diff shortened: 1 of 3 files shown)


wesm:
Good deal. thanks for doing this!

PR #539, opened 2026-03-18
Prevent .git/index.lock contention and enforce read-only sandbox for review agents (#518)

### LINKED ISSUE #518: Constantly running in to .git/index.lock
Ever since I installed roborev on my system I've been *constantly* running into `.git/index.lock` either often existing or being left behind in various repos while I'm trying to stage changes or make commits.

Could this potentially be cause by roborev?

wesm:
## Summary

Fixes #518.

- Set `GIT_OPTIONAL_LOCKS=0` in the environment of all agent subprocesses via `configureSubprocess()`. Git 2.15+ honours this variable and skips the optional index lock that read-only commands (`git status`, `git diff`) normally take to refresh cached stat data. This prevents background agents from contending with the user's own git operations.
- Enforce read-only sandbox for codex review mode by replacing `--full-auto` (which implies `--sandbox workspace-write`) with `--sandbox read-only`. Previously, codex had full write access to the working tree during background reviews — it could run `git add`, modify files, and take mandatory index locks. Now codex review jobs are read-only, matching Claude Code's existing `--allowedTools Read,Glob,Grep` restriction. Agentic mode (fix/refine jobs, `--agentic`) is unchanged and still uses `--dangerously-bypass-approvals-and-sandbox`.
- Use `cmd.Environ()` instead of `os.Environ()` when building subprocess environments so `PWD` is correctly synthesized from `cmd.Dir` rather than inheriting the daemon's stale working directory.
- Move `configureSubprocess()` after Claude's custom env setup so `GIT_OPTIONAL_LOCKS=0` is appended to the final environment rather than being overwritten.
- Probe `codex exec --help` (not top-level `--help`) for `--sandbox` support to match the subcommand where the flag is actually used.

### DIFF
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -344,7 +344,7 @@ func TestSmartAgentReviewPassesModelFlag(t *testing.T) {
 			helpOutput := ""
 			jsonOutput := `{"type": "result", "result": "review result"}`
 			if tt.name == "codex" {
-				helpOutput = "--sandbox --ask-for-approval"
+				helpOutput = "--sandbox"
 				jsonOutput = `{"type": "item.completed", "item": {"type": "agent_message", "text": "review result"}}`
 			}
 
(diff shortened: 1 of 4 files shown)


wesm:
I tested this locally and it appears to be working smoothly.

PR #543, opened 2026-03-19
Check agentsview version before invoking token-use

wesm:
## Summary

- Gate `agentsview token-use` calls behind a version check (`agentsview version` >= 0.15.0) to prevent older installations from spawning a server/opening a browser
- Cache the version probe result per-process using a tri-state (`unchecked`/`ok`/`too_old`) so transient failures (missing binary, exec timeout) are retried on subsequent calls while definitive results are cached permanently
- Add tests for version parsing, old-version skipping, transient-failure retry, and permanent too-old caching


wesm:
Accepted risk

PR #548, opened 2026-03-20
Refactor storage job hydration and verdict parsing helpers

mariusvniekerk:
## Summary
- Extracted job/review scan hydration logic into `internal/storage/hydration.go` to reduce duplication in `jobs.go`.
- Moved verdict parsing and conversion helpers into `internal/storage/verdict.go` with shared pass/fail constants.
- Added focused helper functions for list-job option collection and SQL filter clause construction in `jobs.go`.
- Kept behavior aligned while simplifying `ClaimJob` and related storage-layer mapping code.

## Testing
- Not run (tests were not executed in this change set).
- Verified by inspection that the refactor preserves existing field mapping and verdict-detection paths.

### DIFF
--- a/internal/storage/db_filter_test.go
+++ b/internal/storage/db_filter_test.go
@@ -683,33 +683,6 @@ func TestListJobsUsesStoredVerdictBoolWhenPresent(t *testing.T) {
 	assert.Equal(t, "F", *jobs[0].Verdict)
 }
 
-func TestListJobsUsesStoredVerdictBoolWithEmptyOutput(t *testing.T) {
-	db := openTestDB(t)
-	defer db.Close()
-
-	repo := createRepo(t, db, filepath.Join(t.TempDir(), "empty-output-verdict-repo"))
-
-	job, err := db.EnqueueJob(EnqueueOpts{RepoID: repo.ID, GitRef: "emptyout123", Agent: "codex"})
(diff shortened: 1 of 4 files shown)


wesm:
"⏺ All three findings are invalid.

  Finding 1 (verdict.go:35 — empty output guard): This is the same finding we already addressed in the
  previous round. We intentionally restored the output == "" guard because historical rows have false
  verdict_bool=0 from the old CompleteJob that parsed empty strings as "fail". The CompleteJob fix ensures
  new empty-output rows store NULL. The guard is correct.

  Finding 2 (verdict.go:62 — bare "pass" token): hasSeverityLabel() runs before the pass-phrase loop and
  returns verdictFail if any findings exist. So a review with actual severity labels can't be overridden by
  a stray "pass" line. The bare "pass" match only fires when there are no severity markers at all — it's
  intentional for structured agent output like Verdict: pass (after field-label stripping). The code comment
   at line 50-55 already explains why we don't try to get more clever here.

  Finding 3 (repos.go:441 — partially populated ReviewJob): The premise is wrong. IsTaskJob() depends on
  JobType, CommitID, DiffContent, and GitRef. It does not depend on review_type. All four fields it needs
  are populated by the query via reviewJobScanFields and the GitRef local variable."

PR #514, opened 2026-03-15
Add roborev insights command for review pattern analysis

darrenhaas:
## Summary

- Adds `roborev insights` command that queries failing reviews from the database, builds a structured analysis prompt including current review guidelines, and sends it to an LLM agent for pattern identification (issue #359)
- Agent analyzes recurring finding patterns, hotspot areas, noise candidates, guideline gaps, and produces concrete guideline text suggestions for `.roborev.toml`
- Supports `--repo`, `--branch`, `--since` (default 30d), `--agent`, `--model`, `--reasoning`, `--wait` (default true), and `--json` flags
- Prioritizes unaddressed (open) findings over addressed (closed) ones when truncating to fit prompt size limits
- New files: `cmd/roborev/insights.go`, `internal/prompt/insights.go`, plus tests for both

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #554, opened 2026-03-20
Follow up to #514: make insights a daemon-owned job type

mariusvniekerk:
Follow-up to #514.

Because #514 is opened from a cross-repo branch (`darrenhaas:darren/insights-command`), GitHub cannot stack this PR directly on that branch inside `roborev-dev/roborev`. This PR carries the follow-up work against `main` and references #514 for context.

## Meaningful changes

- add a first-class `insights` job type so the daemon, worker, storage layer, and TUI all treat insights runs as stored-prompt task work
- move insights dataset assembly into the daemon enqueue path, including history selection, branch filtering, review/comment loading, and prompt construction
- simplify the CLI to a thin enqueue client and remove the CLI-side N+1 `/api/jobs` + `/api/review` + `/api/comments` fan-out
- keep review comments in the analysis corpus, exclude compact jobs from the insights dataset, and use the requested branch for insights filtering/metadata
- return a clean skipped response when no matching failing reviews exist instead of forcing the CLI to prefetch history
- add daemon, CLI, and storage tests that lock in the new insights job contract

## Validation

- `go fmt ./...`
- `go vet ./...`
- `go test ./...`

### DIFF
--- a/internal/prompt/insights.go
+++ b/internal/prompt/insights.go
@@ -34,9 +34,9 @@ Identify files, packages, or directories that concentrate failures. List them wi
 
 ## 3. Noise Candidates
 
-Identify finding types that are consistently present in reviews that were closed/addressed without corresponding code changes. These suggest the review guideline should suppress them. For each:
+Identify finding types that are consistently present in reviews that were closed (marked "addressed/closed") and accompanied by developer comments suggesting the finding was intentional or a false positive. These suggest the review guideline should suppress them. For each:
 - Describe the finding type
-- Note how many times it appeared and was dismissed
+- Note how many times it appeared and was closed with dismissive comments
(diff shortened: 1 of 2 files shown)


wesm:
"Technically valid but overstated. The fixed scaffolding (system prompt ~1.5KB + capped guidelines
  ≤1KB + headers) is bounded and small. The per-review budget check on line 183 is where the real
  enforcement happens, and it works correctly.

  The only scenarios where the prompt exceeds MaxPromptSize:
  1. Budget is set smaller than the fixed scaffolding (~2.5KB) — the prompt would be useless anyway
  2. The omission/note lines at the end (~100 bytes) — negligible

  I'd call this a false positive. Not worth adding complexity to guard against configurations that
  would make insights non-functional regardless."

PR #558, opened 2026-03-20
Harden Codex large-diff prompt budgeting and fallbacks

wesm:
## Summary
- harden oversized Codex review prompts so fallback instructions preserve review filtering and stay within prompt budgets
- preserve the legacy prompt-size default, add UTF-8-safe truncation, and enforce a final hard cap for tiny configured budgets
- avoid eagerly materializing oversized commit and range diffs by using bounded git reads
- expand regression coverage for large diffs, large metadata, exclude handling, and small prompt caps

### DIFF
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -397,7 +397,7 @@ func codexCommitInspectionFallbackVariants(sha string, pathspecArgs []string) []
 			"- `%s`\n"+
 			"- `git show %s -- path/to/file`\n\n"+
 			"Review the actual diff before writing findings.\n",
-			statCmd, diffCmd, filesCmd, sha),
+			statCmd, diffCmd, filesCmd, shellQuote(sha)),
 		fmt.Sprintf("### Diff\n\n"+
 			"(Diff too large to include inline)\n\n"+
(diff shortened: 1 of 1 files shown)


wesm:
"Both findings are speculative edge cases, not practical bugs:

  1. BuildDirty with no diff section: If guidelines alone exceed the cap, the diff section
  gets the truncation treatment — same as every other path. The hardCapPrompt is a safety net,
   not the primary trimming mechanism. And in practice, guidelines that exceed 200KB are
  pathological.
  2. Truncation at markdown boundaries: Every truncation in every LLM prompt system works this
   way. Agents handle malformed markdown gracefully — they're not parsers. Engineering
  section-boundary-aware truncation is significant complexity for zero practical benefit.

  These are the kind of findings that sound rigorous but don't correspond to real failure
  modes"

wesm:
"  1. Lockfile excludes in fallback: Same as review 10768. Already decided this is intentional — the agent
  can use judgment. Not changing.
  2. Markdown injection via refs: The sha and rangeRef values in the Codex fallback are already passed
  through shellQuote/renderShellCommand which wraps them in single quotes — those don't contain backticks.
  The non-Codex "View with:" lines aren't inside inline code spans (no backticks), so there's nothing to
  break out of. Not a real issue."

PR #559, opened 2026-03-22
Fix review agent and hook working directory for worktree commits

ryan-mahoney:
## Summary

When commits are made in git worktrees, roborev resolved the path to the main repository root before passing it to review agents and hooks. This caused two problems: review agents searched the main checkout (wrong branch) instead of the worktree, producing false positives for files that only existed on the feature branch; and post-review hooks like `roborev refine` ran in the main repo root, refusing to operate because it detected the `main` branch instead of the feature branch.

- Adds `worktree_path` column to `review_jobs` to persist the originating worktree path per job
- Detects worktree divergence in `handleEnqueue` by comparing `GetRepoRoot` vs `GetMainRepoRoot`
- Worker uses `WorktreePath` as the agent's working directory and prompt builder path, with fallback to the main repo root if the worktree no longer exists
- Hooks now run in the worktree directory, so `roborev refine` operates on the correct branch
- Config resolution, event broadcasting, and fix-job worktree creation continue to use the main repo root

## Changes

- **Add WorktreePath field to ReviewJob and EnqueueOpts**
- **feat(storage): add worktree_path column migration (#1)**
- **Persist WorktreePath in EnqueueJob INSERT and returned struct (#1)**
- **Read worktree_path in ClaimJob, GetJobByID, and ListJobs (#1)**
- **Detect worktree and pass WorktreePath in handleEnqueue (#1)**
- **feat(daemon): use worktree path for agent cwd and prompt building (#1)**
- **test(storage): verify ReenqueueJob preserves worktree_path (#1)**
- **feat(daemon): use worktree path for hook working directory (#1)**
- **fix(daemon): address PR review feedback for worktree path handling (#1)**


ryan-mahoney:
> ## roborev: Combined Review (`9f1246e`)
> **Verdict:** The PR adds `worktree_path` persistence but contains a medium-severity bug affecting directory resolution for non-worktree users.
> 
> ### Medium
> * **Location**: `internal/daemon/server.go:710`
> * **Problem**: `worktreePath` is set whenever `gitCwd != repoRoot`, which also happens when the CLI is run from a subdirectory inside a normal checkout. In that case the code stores the subdirectory path as
>   `WorktreePath`, and later the worker and hook runner use that subdirectory as the review cwd. That changes behavior for non-worktree users and can break prompt-building or hooks that expect the repository root.
> * **Fix**: Resolve and compare the actual checkout root for the current worktree against the main
>   repo root, and only persist `WorktreePath` when those two roots differ. Add a test that enqueues from a repo subdirectory and verifies no worktree path is stored.
> 
> _Synthesized from 3 reviews (agents: codex, gemini | types: default, security)_

No, this specific bug report is not true for the current code.

- In [`handleEnqueue`](\/roborev\/internal\/daemon\/server.go:696), `gitCwd` is computed with `git.GetRepoRoot(req.RepoPath)`, which resolves to the checkout root (not the caller’s subdirectory).
- `repoRoot` comes from [`git.GetMainRepoRoot`](\/roborev\/internal\/daemon\/server.go:704), and for normal repos it falls back to `GetRepoRoot` (same value) in [`internal/git/git.go`](\/roborev\/internal\/git\/git.go:287).
- The CLI already normalizes to repo root before enqueueing in [`review.go`](\/roborev\/cmd\/roborev\/review.go:82) and sends that root as `RepoPath` in [`review.go`](\/roborev\/cmd\/roborev\/review.go:278).

So “run from a subdirectory in a normal checkout stores subdirectory as `WorktreePath`” does not match this implementation. 

The suggested regression test is still a reasonable hardening test, but this specific medium-severity finding appears to be a false positive.

wesm:
False positive. Will merge this once CI green

ryan-mahoney:
Thank you @wesm - I've been enjoying using this service and was delighted to be able to contribute

Issue #556, opened 2026-03-20
OpenCode agent hardcodes obsolete MiniMax model


SaintRod:
Perhaps related to #259 

version: `roborev v0.48.0`

config:
```
default_agent = "opencode"
default_model = "opencode/nemotron-3-super-free"
```

restarted shell via `source ~./bashrc`
restarted rborev daemon via `rborev daemon restart`

`roborev tui`
- pick a job
- rerun via `r`
- job failed
- view logs via 'l'

```
Log #4 1932cd8 (opencode: minimax-m2.5-free) ● complete
Command: opencode run --format json --model minimax-m2.5-free
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
ProviderModelNotFoundError: ProviderModelNotFoundError
 data: {
  providerID: "minimax-m2.5-free",
  modelID: "",
  suggestions: [],
},

      at getModel (/$bunfs/root/src/index.js:185930:15)
```

However, manually specifying a model seems to work
`roborev fix --model opencode/nemotron-3-super-free`

In case it's helpful.
OS = Linux Mint 22.3

mariusvniekerk:
Rerun will use the stored resolved model.  If we want to support this we need to store the intent which in this case was likely `NULL` as the model.  Giving this approach a crack.,

PR #581, opened 2026-03-27
fix: preserve requested model provenance on rerun

mariusvniekerk:
## Summary
- add `requested_model` / `requested_provider` alongside effective `model` / `provider`
- recompute effective model/provider on rerun from requested intent plus current config/defaults
- expose requested model/provider as optional hidden queue columns in the TUI
- extend SQLite, PostgreSQL, and sync paths for the new provenance fields

## Why
Issue #556 came from reruns replaying stale effective OpenCode model values. The old schema only stored the effective model/provider, so rerun could not tell whether a value was explicitly requested or merely resolved from config/defaults.

## Ambiguous migration choice
Existing rows are ambiguous: the historical `model` / `provider` values do not tell us whether they were explicitly requested or implicitly resolved.

I considered storing a per-row creation/model-schema version marker so rerun could special-case older rows, but that adds more schema and branching complexity than the problem warrants.

This change instead chooses the simpler behavior:
- migrate existing rows with `requested_model = NULL` and `requested_provider = NULL`
- do **not** backfill requested fields from historical effective values
- treat `NULL` requested fields as "no explicit request recorded; reevaluate"

That means pre-migration rows will rerun against current defaults/config rather than freezing an old ambiguous effective value.

## Validation
- `go test ./internal/storage ./internal/daemon ./cmd/roborev/tui`
- `go vet ./...`

## Exact scenario checked
I also re-ran the exact configuration discussed in #556:
- `default_agent = "opencode"`
- `default_model = "opencode/nemotron-3-super-free"`

Before this change, that config stored the resolved effective model in `review_jobs.model`, which rerun then replayed as if it were original intent. After this change, explicit requested intent is tracked separately, and config-derived runs leave requested fields empty so rerun can reevaluate.

### DIFF
--- a/internal/daemon/server.go
+++ b/internal/daemon/server.go
@@ -644,9 +644,10 @@ func (s *Server) writeInternalError(w http.ResponseWriter, msg string) {
 
 func workflowForJob(jobType, reviewType string) string {
 	// "default" uses the standard "review" workflow; others use their own name.
-	// Compact jobs use the "fix" workflow since they're part of that pipeline.
+	// Fix and compact jobs use the "fix" workflow since they're part of
+	// that pipeline.
 	workflow := "review"
-	if jobType == storage.JobTypeCompact {
+	if jobType == storage.JobTypeFix || jobType == storage.JobTypeCompact {
 		return "fix"
 	}
 	if !config.IsDefaultReviewType(reviewType) {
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
@wesm I think we probably don't want to do that much backwards compat support here since we'll have to guess either way the upgrade goes.  This path makes old records behave like they do now.

wesm:
As intended.

PR #584, opened 2026-03-27
Add support for systemd socket activation

### LINKED ISSUE #569: Support for systemd socket activation
On Linux systems it seems sensible to run the daemon as a systemd user service for local reviews, but right now there is an awkward choice between starting the service at login (which is almost certainly too early) or making it completely manual.

I'd suggest adding support for systemd socket activation in the daemon so that it can be configured to start up automatically when there is review work to do. I believe there is fairly good support in the Go ecosystem for detecting when the right systemd-related environment variables are set and converting the passed file descriptor to something you can `Listen()` on already.

I'm not sure you'd even need a new flag; it would make sense to me that socket activation takes precedence over `--server` and `--addr` when detected.

atheriel:
This commit facilitates running roborev as a socket-activated systemd user service. When `LISTEN_FDS` is set (handled by go-systemd internals), the daemon uses the systemd-provided socket instead of creating its own, enabling on-demand startup and lifecycle management driven by systemd.

We enforce that socket activation is used only with loopback TCP listeners and unix sockets with safe permissions.

We also avoid cleaning up the socket file in this mode, since we don't own it.

Closes #569.

Sample systemd user service and socket files, which we could put into the documentation:

```
[Unit]
Description=roborev daemon
Requires=roborev.socket

[Service]
Type=simple
ExecStart=%h/.local/bin/roborev daemon run
Restart=on-failure

[Install]
WantedBy=default.target
```

```
[Unit]
Description=roborev daemon socket

[Socket]
ListenStream=%t/roborev/daemon.sock
SocketMode=0600

[Install]
WantedBy=sockets.target
```

These both go into `~/.config/systemd/user`. To run `roborev` as a socket-activated user service, you'd then:

```bash
$ systemctl --user daemon-reload
$ systemctl --user enable --now roborev.socket
# Use roborev, e.g.:
$ roborev --server "unix://$XDG_RUNTIME_DIR/roborev/daemon.sock" status
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
 
             src = ./.;
 
-            vendorHash = "sha256-3eyTM8oYEkmAHshFGDTrbVWU106zvP48nDnrGtAta9M=";
+            vendorHash = "sha256-PE3kbfJQlvUeSPmLawxtVnqTEz+6EI+TS8dc7jphl7w=";
 
             subPackages = [ "cmd/roborev" ];
 


wesm:
Addressed some minor things from running roborev reviews, will merge this soon

PR #636, opened 2026-04-09
Prompt agents to read project instruction files during reviews

shoyer:
## Summary

- Adds an instruction to the review prompt (`noSkillsInstruction`) telling agents to read project instruction files (e.g., CLAUDE.md, AGENTS.md) from the repo root and from directories containing changed files before reviewing.
- I've been seeing review agents clearly not following guidance from my CLAUDE.md files. In print mode (`-p`), agents focus on responding to the provided prompt and don't proactively explore the filesystem for instruction files the way they would in an interactive session. This is the simplest fix I could think of — just tell them to look. Open to other approaches if there's a better way.

### DIFF
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -34,7 +34,7 @@ IMPORTANT: You are being invoked by roborev to perform this review directly. Do 
 Return only the final review content. Do NOT include process narration, progress updates, or front matter such as "Reviewing the diff..." or "I'm checking...".
 If you use tools while reviewing, finish all tool use before emitting the final review, and put the final review only after the last tool call.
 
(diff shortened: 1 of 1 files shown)


wesm:
Aside from the spurious security concerns, does this actually improve behavior conformance? I would think that `codex exec` or `claude -p` in the repository would automatically read the AGENTS.md/CLAUDE.md

shoyer:
I think the problem is that reviewer agents will see diffs, but not necessarily read CLAUDE.md files that cover the files in those diffs.

I haven't tested this new prompt enough yet to know if it works. But certainly I've found it really annoying to be working on a project where my CLAUDE.md says "this is experimental code, don't worry about backwards compatibility" and every roborev review says "add backwards compat!"

wesm:
are you using `review_guidelines`? https://www.roborev.io/configuration/#review-guidelines

wesm:
I think this is best solved via the review guidelines in the review prompt so closing this in favor of that

PR #640, opened 2026-04-12
fix: allow internal whitespace in severity threshold marker for Gemini agent

graycoldknight:
Fixes an issue where the Gemini model agent sometimes outputs the `SEVERITY_THRESHOLD_MET` marker with internal newlines (e.g., `SEVERITY_THRESHOLD\n_MET`), causing the `IsMarkerOnlyOutput` check to fail and incorrectly flag the review as a failure.

### Evidence
We tested this by feeding `SEVERITY_THRESHOLD\n_MET` and `SEVERITY_THRESHOLD\n\n_MET` to `IsMarkerOnlyOutput`. Previously it failed, but with the stripping of internal whitespace (`\n`, `\r`, `\t`, ` `) before the final string comparison, it passes correctly. We've added these cases to the test suite in `internal/config/config_test.go`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1493,7 +1493,12 @@ func IsMarkerOnlyOutput(output string) bool {
 	// Strip a trailing period.
 	s = strings.TrimSpace(strings.TrimSuffix(s, "."))
 
-	return s == SeverityThresholdMarker
+	// Collapse any internal whitespace models might inject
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
We should probably make this conditional on the gemini models

wesm:
Seems harmless enough. Merging

PR #656, opened 2026-04-19
tui: fix clipboard over SSH using OSC52 escape sequences

lgonzalezsa:
## Problem

When `roborev tui` is used over an SSH connection, pressing `y` to copy a review shows "Copied to clipboard ✓" but the clipboard is empty on the local machine.

**Root cause:** `realClipboard.WriteText()` calls `xclip`/`xsel`/`wl-clipboard` on the *remote* machine. This only reaches the user's local clipboard if X11 forwarding is active. Terminals that don't forward X11 (e.g. Tabby's built-in SSH plugin) silently lose the data.

## Fix

Add an `osc52Clipboard` implementation that writes the [OSC52 terminal escape sequence](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) (`\x1b]52;c;<base64>\x07`) to stderr. The sequence travels through the SSH pipe and is intercepted by the **local** terminal emulator, which sets the system clipboard directly — no X11 forwarding required.

`newClipboard()` auto-selects the implementation:
- **SSH session** (`SSH_TTY` / `SSH_CLIENT` / `SSH_CONNECTION` env var is set) → `osc52Clipboard`
- **Local session** → `realClipboard` (unchanged behaviour)

## Compatibility

OSC52 is supported by all modern terminals: Alacritty, Tabby (xterm.js), Kitty, WezTerm, iTerm2, Windows Terminal, and more.

## Changes

- `cmd/roborev/tui/types.go` — add `osc52Clipboard` struct; promote `go-osc52/v2` from indirect to direct dependency
- `cmd/roborev/tui/tui.go` — add `newClipboard()` factory with SSH detection; wire into `newModel`
- `cmd/roborev/tui/review_clipboard_test.go` — 5 new tests: OSC52 output format, empty string, SSH env var detection (SSH_TTY / SSH_CLIENT / SSH_CONNECTION / no-SSH)

### DIFF
--- a/cmd/roborev/tui/review_clipboard_test.go
+++ b/cmd/roborev/tui/review_clipboard_test.go
@@ -1,6 +1,7 @@
 package tui
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"testing"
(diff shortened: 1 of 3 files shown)


wesm:
Inflated severity, this looks fine

PR #660, opened 2026-04-19
fix: use branch upstream (not default branch) for --branch reviews

cpcloud:
## Summary

- `review --branch`, `tryBranchReview`, `analyze --branch`, and `refine` resolved the merge-base against the repository's default branch (typically `origin/main`). In fork workflows where the current branch's actual upstream (e.g. `upstream/main`) is ahead of `origin/main`, this pulled in commits that were already merged upstream. Each command now prefers `@{upstream}` of the current branch and falls back to `GetDefaultBranch()` only when no upstream is configured.
- Added `git.GetUpstream(repoPath, ref)` (returns the upstream tracking branch via `git rev-parse --abbrev-ref <ref>@{upstream}`, empty on no upstream).
- Replaced the `currentBranch == LocalBranchName(base)` guardrails with `git.IsOnBaseBranch(repoPath, currentBranch, base)`, which generalizes the `origin/` shortcut by verifying `refs/remotes/<base>` exists before stripping the remote prefix — so non-origin remotes are handled and local branches with slashes (e.g. `feature/foo`) are not misclassified.
- New tests: `TestGetUpstream`, `TestIsOnBaseBranch`, and `TestTryBranchReview` cases for the non-origin upstream scenario and for the feature/foo misclassification guard.

### DIFF
--- a/cmd/roborev/analyze.go
+++ b/cmd/roborev/analyze.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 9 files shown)


cpcloud:
@wesm this is ready to go

PR #661, opened 2026-04-19
feat: opt-in automatic design-review router

cpcloud:
## Summary

- Adds an opt-in per-commit decision (heuristics + schema-constrained classifier fallback) that routes design reviews automatically; off by default.
- New `internal/review/autotype` package: doublestar path globs, diff-size thresholds, commit-subject regexes, with trigger-before-skip-before-classifier precedence.
- New `SchemaAgent` agent capability returning JSON conforming to an embedded schema via `claude-code --json-schema` and `codex exec --output-schema`; non-schema agents rejected at config-resolve time.
- New `JobStatusSkipped` terminal state and `JobTypeClassify`, plumbed through every existing terminal-state filter (`GetStaleBatches`, `ReconcileBatch`, sync, `ReenqueueJob`, `HasMeaningfulBatchResult`, `GetExpiredBatches`).
- New `skip_reason` and `source` columns on `review_jobs` plus two partial unique indexes scoped to `source='auto_design'` (commit-backed and ref-only); SQLite migration + Postgres v12 schema mirror.
- Classify-row lifecycle helpers (`PromoteClassifyToDesignReview`, `MarkClassifyAsSkippedDesign`, `InsertSkippedDesignJob`, `EnqueueAutoDesignJob`) — classify rows convert in place via UPDATE so the dedup index isn't fought.
- Daemon integration: worker handler for `classify` jobs; enqueue-handler dispatches the auto-design path for default-typed `review`/`range`/`dirty` jobs only; CI poller runs the same path on the PR head SHA when `design` isn't already in the matrix; heuristic-input failures degrade to classifier (not skip).
- TUI renders skipped design rows with a dimmed style and truncated reason; PR synthesis includes a short `Auto-design-review skipped: <reason>` section; `/api/status` exposes a `SkippedJobs` aggregate count and, when the feature is enabled anywhere, an `auto_design` subobject with five per-outcome counters. Subobject omitted from JSON when disabled.
- Heuristic-engine benchmark on the spec's target hardware: trigger 113 ns/op, skip 5 µs/op, ambiguous 7.7 µs/op (spec target: < 25 ms).
- Spec at `docs/superpowers/specs/2026-04-17-auto-design-review-design.md`, plan at `docs/superpowers/plans/2026-04-17-auto-design-review.md`, user docs at `docs/auto-design-review.md`.


cpcloud:
Re: roborev-ci review on `217ec60` ([comment](https://github.com/roborev-dev/roborev/pull/661#issuecomment-4317663552)) — the two High findings are not real bugs. The Medium is, and is fixed in `052a866`.

### High #1 — `parent.CommitIDValue()` "doesn't exist, will fail to compile"

The method is defined at [`internal/storage/models.go:110-112`](https://github.com/roborev-dev/roborev/blob/217ec60/internal/storage/models.go#L110-L112):

```go
// CommitIDValue returns the commit ID as a plain int64 (0 if nil).
func (j ReviewJob) CommitIDValue() int64 {
    ...
}
```

`go build ./...` at `217ec60` exits 0; the CI `test` jobs on this commit (ubuntu/macos/windows) all passed.

### High #2 — `ensureCIPRBatchJobsUniqueIndex` "references `id`, but `ci_pr_batch_jobs` has no `id` column"

The schema at [`internal/storage/db.go:107-112`](https://github.com/roborev-dev/roborev/blob/217ec60/internal/storage/db.go#L107-L112):

```sql
CREATE TABLE IF NOT EXISTS ci_pr_batch_jobs (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  batch_id INTEGER NOT NULL REFERENCES ci_pr_batches(id),
  job_id INTEGER NOT NULL REFERENCES review_jobs(id),
  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```

The `id` column exists. `go test ./internal/storage/...` (which calls `OpenDB`, which calls `ensureCIPRBatchJobsUniqueIndex`) passes locally and in CI.

### Medium — stuck `running` classify row on DB transition error

This one is real. `applyClassifyVerdict` and `completeClassifyAsSkip` logged the `PromoteClassifyToDesignReview` / `MarkClassifyAsSkippedDesign` error and returned, leaving the row in `running`. Fixed in `052a866`: a shared `failClassifyOnDBError` helper now calls `FailJob` and broadcasts `review.failed` so any linked CI batch advances instead of waiting for stale-batch reconciliation. Tests in `worker_classify_test.go` force the inner UPDATE's WHERE clause to miss (via `source` mutation) and assert the row transitions to `failed` with `review.failed` broadcast.

cpcloud:
This should be ready to go.

wesm:
Ok, firing torpedoes!

PR #670, opened 2026-04-26
migrate daemon API to huma

mariusvniekerk:
## Summary
- consolidate daemon routes/types/handlers into Huma-backed server files
- add generated internal daemon client from Huma OpenAPI 3.0 output
- move TUI normal JSON calls onto the generated daemon client

## Validation
- go generate ./internal/daemon_client
- go test ./...
- go vet ./...


wesm:
Is this mergeable?

mariusvniekerk:
Should be, main decisions are if we want to check in the generated openapi.json/yaml schema?  Probably useful for other tools to build upon

wesm:
Seems okay to me, it will generate some churn but the API isn't changing that often now

PR #671, opened 2026-04-26
feat(tui): add per-severity finding counts to queue, review, and tasks views

cpcloud:
## Summary

- Add `CountFindings` helper that extracts high/medium/low counts from review output (critical+high → H, medium → M, low+info → L) and persist them as new columns on the `reviews` table in both SQLite and PostgreSQL (Postgres bumped to schema v13).
- Backfill existing rows on schema upgrade in both databases; sync workers recompute counts from review text on pull so legacy zero rows get populated rather than overwriting freshly counted values.
- Surface counts in the TUI: new `Findings` column in the queue table (toggleable, persisted in column config), aggregate `Findings: H<n> M<n> L<n>` line in queue status header, finding badges in the review detail header, and parent-review counts in the tasks view.
- Severity badge styling uses red (high), yellow (medium), and blue (low) with light/dark-theme variants; zero counts render dim. Nil pointers are guarded throughout via a shared `derefOrZero` helper.
- `JobStats` carries finding counts so `roborev status` reflects them; `ReviewJob` carries both own and parent counts so the tasks view can display the parent review's findings without a second fetch.

### DIFF
--- a/cmd/roborev/tui/queue_test.go
+++ b/cmd/roborev/tui/queue_test.go
@@ -598,9 +598,11 @@ func TestTUIJobCellsContent(t *testing.T) {
 		job.Closed = &handled
 
 		cells := m.jobCells(job)
+		// cells: ref(0), branch(1), repo(2), agent(3), queued(4),
+		// elapsed(5), status(6), pf(7), findings(8), handled(9), session(10), ...
 		assert.Equal(t, "Done", cells[6])
 		assert.Equal(t, "P", cells[7])
-		assert.Equal(t, "yes", cells[8])
+		assert.Equal(t, "yes", cells[9])
 	})
 }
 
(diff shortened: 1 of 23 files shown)


cpcloud:
Demo of the new `Findings` column with the tightened `3/2/5` badge format.

**Queue view** — status header aggregates findings across visible jobs (`Findings: 3/6/11`), per-row badges show each review's own counts. Slashes and zeros render dim grey so non-zero counts pop in red (high), yellow (medium), blue (low).

roborev queue

**Review detail** — header line shows `Verdict: ...  Findings: 0/2/5` for the selected review.

roborev review

wesm:
I wonder if the high/medium/low distinction survives language changes (e.g. if someone asks for the results to be in German)

cpcloud:
Good question, I doubt it. Seems like to account for that there would have to be a deeper refactor to have agents produce structured output with English-based JSON keys, and construct the TUI/UX from that instead of the current approach. I think this would be a backwards compatible change, only internals would see the JSON for now.

cpcloud:
I'll get clanking on that, and we can wait on this PR until that is in.

cpcloud:
Well, basically most of the supported review agent CLIs do not support anything like `--output-schema` (claude code and codex do, gemini has an issue for it), so we're back to the same problem if we actually want to count these across languages. Kind of unfortunate given that many of them support structured outputs at the API layer. They just don't expose it on the CLI.

This PR can be closed out if we think it's too fragile or specific to English-output review text.

mariusvniekerk:
i think we can keep it in draft until we come up with a better design

PR #677, opened 2026-04-29
Fix detached-HEAD review discovery in roborev fix

mariusvniekerk:
## Summary
- Teach open-job filtering to fall back to detached commit refs when no branch is checked out, so `roborev fix` can still find reviews in detached HEAD state.
- Match jobs against the detached commit chain and range endpoints, while keeping `dirty` and unrelated refs excluded.
- Add regression coverage for detached HEAD discovery and the surrounding filtering cases.

## Testing
- `go fmt ./...`
- `go vet ./...`
- `go test ./...`

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -515,9 +515,17 @@ func filterReachableJobs(
 	if matchBranch == "" {
 		matchBranch = git.GetCurrentBranch(worktreeRoot)
 	}
+	var detachedRefs map[string]struct{}
+	if matchBranch == "" {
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
Addressed the detached merge checkout case in 966677a. The first-parent-only walk was narrow enough for linear detached chains, but it could miss jobs whose ref was on another parent of a detached merge commit, so the detached fallback now traverses all direct parents until each path reaches a branch boundary.

Also added regression coverage for the detached merge case and direct parent helper coverage. Local validation: `go fmt ./...`, `go vet ./...`, `go test ./...`.

PR #683, opened 2026-05-03
Fix tmux terminal probe leakage in non-TUI commands

wesm:
## Summary

Fixes terminal escape leakage like `^[]11;rgb:1515/1111/1010^[\` and `^[[18;91R` that appears in the parent shell when roborev runs inside tmux — most visibly when the post-rewrite git hook fires `roborev remap --quiet` during `git rebase`.

### Root cause

`charmbracelet/bubbletea` calls `lipgloss.HasDarkBackground()` from its package `init()` to work around a separate hang. That cascades into `termenv.termStatusReport`, which writes OSC 11 + cursor-position queries to the TTY. The single roborev binary imports the TUI package transitively from many subcommands, so this probe runs on every invocation — including non-TUI ones like `roborev remap --quiet` and `roborev --help`. tmux normally swallows the response, but in nested setups the bytes leak to the parent shell.

Upstream termenv already short-circuits when `TERM` starts with `screen`/`tmux`/`dumb`, but the common true-color tmux setup advertises `TERM=xterm-256color`, so that check doesn't fire. Checking `$TMUX` (which tmux always exports inside a session) is the missing piece. Kata previously shipped the same fix.

### Why vendor + patch instead of a lighter workaround

I considered three alternatives before vendoring:

1. **Normalize `$TERM` at process start** — set `TERM=tmux-256color` if `$TMUX` is set, before bubbletea's init fires the probe. Doesn't work: bubbletea's package `init()` runs before `cmd/roborev/main.go`'s own `init()` (Go runs deepest-dependency inits first), so we have no opportunity to mutate the environment in-process. We'd need an underscore-imported "earlyinit" package whose dependency tree is guaranteed to initialize before bubbletea's — a constraint that's invisible to readers and one rebase away from breaking.
2. **Hook-level `TERM` rewrite** — add `[ -n "$TMUX" ] && export TERM=tmux-256color` to the generated git hook. Fixes `git rebase` but not `roborev --help` or any other manual non-TUI invocation in tmux, both of which leak today.
3. **Set up a custom `termenv.Output` before bubbletea grabs the default** — same init-order problem as (1).

Vendoring fixes it at the actual bottleneck (`termStatusReport`), which is the single function every probe path goes through. It can't be bypassed by a future dependency update or import reordering, and the patch is a five-line `if` block that's trivial to re-apply on upgrades. `internal/third_party/termenv/PATCHES.md` documents the patch and the upgrade procedure.

### Changes

- Vendor `github.com/muesli/termenv` v0.16.0 into `internal/third_party/termenv` and add the `$TMUX` guard at the top of `termStatusReport`.
- Wire it in via `replace github.com/muesli/termenv => ./internal/third_party/termenv` in `go.mod`.
- Update `flake.nix` `vendorHash` to match the new `goModules` output (the replace target is now part of the vendored tree).
- Bump generated git hooks to redirect stdout to `/dev/null` (`>/dev/null 2>/dev/null`) as a belt-and-suspenders mitigation in case a future dep adds another probe path. Hook version markers bumped to `v5`/`v3` so existing installs auto-upgrade.
- Add unit tests in the vendored package proving the guard fires under tmux, plus an integration test in `internal/streamfmt` that exercises the patched `termenv` through the main module's `replace` directive (verified to fail when the patch is removed).


wesm:
Closing this without merging.

The escape leakage we were patching around is not actually a bug in roborev — it's an observed bug in **superset**, where the tmux layer is failing to consume terminal probe responses (OSC 11, CSI 6n) that should be silently absorbed. tmux normally swallows these transparently for every tool that imports `bubbletea`/`termenv`/`lipgloss` (lazygit, gh, fzf, k9s, etc.), and roborev is no different — it's just one of many victims that exposed the regression.

Patching `termenv` here would only stop the noise for roborev, leaving every other terminal tool you run in superset still leaking through the same path. The right fix is in superset itself; once that lands, this workaround becomes dead weight on every termenv upgrade.

Tracking the superset-side investigation separately.

PR #693, opened 2026-05-05
feat(fix): add --batch-size N and --resume to roborev fix

wesm:
## Summary

- Add `--batch-size N`: pack up to N reviews into a single agent invocation, still bounded by `max_prompt_size`. Mutually exclusive with `--batch` and `--list`.
- Add `--resume`: reuse the agent's session ID across calls within one fix run. Defaults to off.
- Move `sessionCaptureWriter` from `internal/daemon` to `internal/agent` so both daemon and CLI can wrap streaming output to extract session IDs.
- Teach the `test` agent to mint per-instance session IDs (`test-session-N`) and echo `WithSessionID(id)` so resume chains are deterministic in tests.
- Defer fix-agent resolution until a job actually needs an agent (`ensureBaseAgent`), so no-op paths (no open jobs, all already passing, invalid IDs) don't fail at agent resolution.


cmd/roborev/fix.go
```diff
@@ -195,7 +214,10 @@ Examples:
 	cmd.Flags().BoolVar(&allBranches, "all-branches", false, "include open jobs from all branches")
 	cmd.Flags().BoolVar(&newestFirst, "newest-first", false, "process jobs newest first instead of oldest first")
 	cmd.Flags().BoolVar(&batch, "batch", false, "concatenate reviews into a single prompt for the agent")
+	cmd.Flags().IntVar(&batchSize, "batch-size", 0, "concatenate up to N reviews per agent invocation (cap by count, still bounded by max_prompt_size)")
 	cmd.Flags().BoolVar(&list, "list", false, "list open jobs without fixing")
+	cmd.Flags().BoolVar(&resume, "resume", false, "resume the agent's session ID across calls within this run")
+	cmd.Flags().BoolVar(&noResume, "no-resume", false, "explicitly disable session resume (default off)")
```
mariusvniekerk: these two flags seem to touch the exact same thing?  do we really need both?

cmd/roborev/fix_session.go
```diff
@@ -0,0 +1,95 @@
...
+// fixSessionTracker keeps the most-recent agent session ID across
+// calls within a single roborev fix run and produces the agent to
+// invoke next. See spec for full contract.
+type fixSessionTracker struct {
+	enabled bool         // --resume passed
+	base    agent.Agent  // resolved once (already WithAgentic/WithReasoning/WithModel)
+	last    string       // most recent valid captured session ID
+	warned  bool         // log "agent X doesn't support resume" only once
+	quiet   bool         // suppress progress logs and warnings under --quiet
+	log     func(string) // injected logger; cmd.Printf in CLI, no-op in tests
```
mariusvniekerk: feels a bit weird?  dont we have a proper configured logger?

mariusvniekerk (commented):
Some observations

wesm:
clanking

PR #696, opened 2026-05-05
Abort foreground fix loop on agent quota or session limit

wesm:
## Summary

- Add a rate-limit classifier inside `internal/agent` (`LimitKind`, `LimitClassification`, `LimitClassifier`, `ClassifyLimit`, plus `ParseResetDuration` / `ParseResetTime`). The default ruleset reproduces the nine quota substrings the daemon worker already used; reset-time parsing handles relative (`reset after 48m`) and absolute (`resets at 5:42 PM`) hints, and rolls forward via `time.Date(..., Day()+1, ...)` so wall-clock time is preserved across DST transitions.
- Route the daemon worker's quota detection through `agent.ClassifyLimit` via an injectable `WorkerPool.classify` field. Cooldown duration order is unchanged: `defaultCooldown` → `Classification.CooldownFor` → `time.Until(ResetAt)`. `LimitKindQuota` cools down and skips retries; `LimitKindNone` now logs a one-line WARN with a truncated error preview so unmatched agent errors are visible.
- Add `agentLimitError`, `formatAgentLimitMessage`, and `agentLimitLabel` in `cmd/roborev/fix.go`. The foreground `roborev fix` loop classifies each `fixJobDirect` error and aborts with `agentLimitError` on `LimitKindQuota` / `LimitKindSession`; `LimitKindNone` prints a stderr warning. The abort path is wired into `fixSingleJob`, `runFixBatch`, and `runFixWithSeen` (so discovery mode no longer demotes quota aborts to per-job warnings).
- `fixJobDirect`'s commit-retry path also classifies its own error so a quota or session limit on the second `Agent.Review` call returns `*agentLimitError` instead of being swallowed as a `Warning: commit agent failed:` line. The classifier is plumbed through `fixJobParams.Classify` (defaults to `agent.ClassifyLimit` when nil).
- `LimitKindSession` plumbing (cooldown on the daemon side, abort on the CLI side) is in place but is not produced by any production rule yet — Claude's 5-hour cap falls through to the unmatched-error path until a real session-cap message is captured. Speculative substrings are deliberately not added to `defaultLimitRules` because they would also match policy errors and transient 429s, and a false positive would abort `roborev fix` and cool down the agent when retrying might have worked. Tests exercise the session-limit path via an injected classifier.
- Replace the daemon's `isQuotaError`, `parseQuotaCooldown`, `minCooldown`, and `maxCooldown` with the classifier-backed flow.
- Tighten `.roborev.toml` review guidelines to tell reviewers not to flag suspected compile errors, missing imports, or undeclared identifiers — the local toolchain catches those before merge, and reviewers cannot see the rest of the package.


wesm:
unrelated CI flake

PR #705, opened 2026-05-08
Add generated diff categorization for Linguist and lockfiles

mariusvniekerk:
## Summary
- Add a `generated` auto-design category for diffs that are entirely generated, using `linguist-generated` Git attributes plus built-in generated artifacts like lockfiles.
- Thread generated-file detection through daemon auto-design dispatch so generated-only changes skip with an explicit `generated` reason instead of being treated as trivial or large diffs.
- Add unit and end-to-end coverage for Git attribute detection, generated-only classification, and auto-design behavior.

## Testing
- `go fmt ./...`
- `go vet ./...`
- `go test ./...`
- `git diff --check`


mariusvniekerk:
wrong repo.

PR #707, opened 2026-05-09
feat(tui): hide auto-design-router noise from queue by default

cpcloud:
## Summary

- Adds `show_classify_jobs` config (off by default, global + per-repo `*bool` override) that hides auto-design-router byproducts from the TUI queue: rows with `job_type='classify'` and rows with `status='skipped'`. Both predicates are scoped to `source='auto_design'` so a future pipeline that adopts the same job_type or status for unrelated reasons isn't silently swallowed.
- Server-side filtering via new `/api/jobs?hide_classify_jobs=true` query param backed by `storage.WithHideClassifyJobs`. SQL uses `COALESCE` on `j.source` so rows with `source IS NULL` aren't dropped by NULL three-valued logic.
- Per-repo override applies only when the queue is filtered to a single repo, otherwise the global value applies, avoiding one repo's preference leaking into a multi-repo view.
- Session-only `s` hotkey toggles visibility; the queue footer label and the `?` help view both reflect the current state.
- Hitting `l` on an auto-design classify or terminal skipped row renders the classifier verdict, `skip_reason`, and any internal error in a header band above the (typically empty) streamed log. Reasoning lines are width-truncated and have embedded newlines/tabs folded to spaces so each occupies exactly one terminal row, keeping `logVisibleLines` accurate. Verdict text branches on `job.Status` so failed/canceled classifiers don't claim "in progress".
- Effective show-classify value is cached on the model and recomputed only on filter or session-toggle changes, keeping render and fetch paths free of `LoadRepoConfig` disk I/O.

### DIFF
--- a/cmd/roborev/tui/control_handlers.go
+++ b/cmd/roborev/tui/control_handlers.go
@@ -254,6 +254,7 @@ func (m model) handleCtrlSetFilter(
 	m.fetchSeq++
 	m.queueColGen++
 	m.loadingJobs = true
+	m.recomputeClassifyEffective()
 	return m, controlResponse{OK: true}, m.fetchJobs()
 }
 
@@ -299,6 +300,7 @@ func (m model) handleCtrlClearFilter(
 	m.fetchSeq++
 	m.queueColGen++
 	m.loadingJobs = true
+	m.recomputeClassifyEffective()
 	return m, controlResponse{OK: true}, m.fetchJobs()
 }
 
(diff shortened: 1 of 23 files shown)


cpcloud:
False positive on the latest roborev-ci finding. The review claims `ListJobs` scans nullable `j.skip_reason` and `j.source` directly into `string` fields, which would fail on NULL.

In fact the scan targets are `sql.NullString`:

```go
// internal/storage/hydration.go
type reviewJobScanFields struct {
    ...
    SkipReason        sql.NullString
    Source            sql.NullString
}
```

The call site at `internal/storage/jobs.go:880-887` scans into `&fields.SkipReason, &fields.Source` — `sql.NullString` handles NULL natively. The hydration step then unwraps via `.Valid`:

```go
if fields.SkipReason.Valid {
    job.SkipReason = fields.SkipReason.String
}
if fields.Source.Valid {
    job.Source = fields.Source.String
}
```

`TestListJobsWithHideClassifyJobs` already exercises this with a row whose `source IS NULL` (the test asserts that row remains visible after the hide filter); it passes. No code change.

mariusvniekerk (approved):

PR #721, opened 2026-05-15
prompt: let review agents read repo files to verify diff claims

cpcloud:
## Summary

- Review templates told agents to base findings on "the diff and static analysis only". Agents took that literally and treated anything not in the diff as absent from the repo, producing false-negative claims like "this package doesn't exist" on docs PRs that describe code living outside the diff.
- Drop the "diff only" framing in the codex, claude-code, and gemini review templates. Keep the prohibition on building, running tests, and executing code.
- Explicitly permit reading other repo files to verify claims, and instruct agents to check existence before asserting a referenced file, package, function, or symbol is absent.
- Non-templated agents (copilot, opencode, cursor, kiro, kilo, droid, pi) still hit the fallback constants in `prompt.go` and are unchanged; they aren't currently used for CI review.


wesm (approved):
Good thinking, LGTM

mariusvniekerk (approved):
Yeah :shipit:

PR #723, opened 2026-05-16
fix(agent): adapt Codex resume args

mariusvniekerk:
## Summary
- invoke Codex resumed sessions as `codex exec resume --json ...` without resume-unsupported runtime flags
- preserve read-only review behavior through config overrides for resumed runs
- expose diff snapshot directories to resumed Codex reviews through a temporary permissions profile

## Validation
- `go fmt ./...`
- `git diff --check`
- `go vet ./...`
- `go test ./internal/agent -count=1`
- `go test ./...`

### DIFF
--- a/internal/agent/codex.go
+++ b/internal/agent/codex.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log"
 	"os/exec"
+	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


wesm:
recommend cutting 0.55.1?

PR #727, opened 2026-05-20
fix(cli): show usage for invocation errors, plain Error for runtime failures

cpcloud:
## Summary

Cobra's default behavior dumps the full usage block on every `RunE` error, which buries actual error messages under a wall of help text. The codebase had been working around this by scattering `cmd.SilenceUsage = true` assignments inside individual `RunE` handlers — a whack-a-mole pattern that drifted out of sync over time. This PR centralizes the policy and distinguishes three failure categories:

| Failure | Output |
|---|---|
| Invocation error (unknown flag, bad arg, unknown subcommand, mutex flags, bad enum, missing value, invalid `--server`) | `Error: ...` + usage block |
| Runtime error (daemon down, bad git ref, network failure, polling error) | `Error: ...` only |
| Caller-controlled exit (review verdict-fail, multi-job wait fail) | silent, exit with the right code |

## How it works

- **Root `PersistentPreRunE`** sets `cmd.SilenceUsage = true` **after** `validateServerFlag`. Cobra's invocation validation (flag parsing, arg validation, subcommand match, `--server`) all run while `SilenceUsage` is still false, so usage prints. Anything reached via `RunE` runs with usage silenced.
- **`usageErr(cmd, err)`** re-enables `SilenceUsage` for the specific case of `RunE`-internal flag validation (mutex flags, bad enums, missing required values). Applied across `review`/`wait`/`fix`/`compact`/`comment`/`show`.
- **`silentExit(cmd, code)`** returns an `*exitError` and sets `SilenceErrors` on `cmd`. Used for caller-controlled silent exits in single-threaded contexts.
- **`silenceIfExit(cmd, err)`** does the same conditionally; used at top-level `RunE` returns to silence cobra when the error came from concurrent code that couldn't safely mutate `cmd` itself. `showReview` is called from `waitMultiple` goroutines, so it now returns a bare `*exitError` with no `cmd` mutation; the serial `RunE`/post-join code silences after the fact. Verified clean under `go test -race`.
- **`quietExit(cmd, err)`** is the `--quiet` defer wrapper: silences the usage wall (so an upstream `usageErr` doesn't leave a help block in quiet mode) and wraps plain errors in `exitError{code:1, cause: err}`. It does **not** silence cobra's `Error: ...` line — runtime failures in quiet mode still tell the user what broke, matching CLI convention (`--quiet` < `--silent`). The `exitError` type carries an optional `cause` so the underlying message is preserved through `Error()`/`Unwrap()` for both display and programmatic inspection.
- **post-commit** hook entrypoint sets `SilenceErrors`/`SilenceUsage` on its own struct. It's a git-hook contract — failures go to the hook log file, never to git's stderr.

## Resulting behavior

| Invocation | Output |
|---|---|
| `roborev wait --bogus` | `Error: unknown flag: --bogus` + usage |
| `roborev review --branch --dirty` | `Error: cannot use --branch with --dirty` + usage |
| `roborev review --type=bogus` | `Error: invalid --type "bogus" (valid: security, design)` + usage |
| `roborev foobar` | `Error: unknown command "foobar" for "roborev"` + `Run 'roborev --help' for usage.` |
| `roborev --server=://garbage wait` | `Error: invalid --server address ...` + usage |
| `roborev wait --sha=does-not-exist` | `Error: invalid git ref: does-not-exist` (no usage) |
| `roborev wait --quiet --sha=does-not-exist` | `Error: invalid git ref: does-not-exist` (no usage) |
| `roborev wait --quiet --job` | `Error: --job requires a job ID argument` (no usage) |
| `roborev review --wait` (verdict FAIL) | already-printed review output, exit 1, no `Error:` line |
| `roborev review --wait --quiet` (verdict FAIL) | silent, exit 1 |

## Tests

- The `wait` / `review` test helpers now mirror production by setting `SilenceUsage` on the freshly-constructed subcommand (the root's `PersistentPreRunE` doesn't run for direct-`Execute` tests).
- `TestWaitReviewFetchErrorExposesCauseInQuietMode` replaces `TestWaitReviewFetchErrorIsPlainError`, whose assertion predated this work and contradicted both the current `--quiet` contract and the visible-runtime-errors invariant.
- `TestWaitForJobUnknownStatus` passes now that `exitError` preserves the underlying cause; it was failing at HEAD before this work.

### DIFF
--- a/cmd/roborev/helpers.go
+++ b/cmd/roborev/helpers.go
@@ -9,15 +9,78 @@ import (
 	"github.com/roborev-dev/roborev/internal/git"
 	"github.com/roborev-dev/roborev/internal/githook"
 	"github.com/roborev-dev/roborev/internal/storage"
+	"github.com/spf13/cobra"
 )
 
-// exitError is an error that signals a specific exit code
+// exitError signals a specific exit code with no further error output.
+// Always construct via silentExit so cobra's "Error: ..." line is suppressed
+// for this command — the caller has already printed any user-facing message.
+// If cause is set, Error()/Unwrap() expose it so programmatic callers can
+// still inspect the underlying failure; cobra's printing is still silenced
+// via the SilenceErrors flag that silentExit sets.
 type exitError struct {
-	code int
+	code  int
+	cause error
 }
 
 func (e *exitError) Error() string {
+	if e.cause != nil {
+		return e.cause.Error()
+	}
 	return fmt.Sprintf("exit code %d", e.code)
+}
+
+func (e *exitError) Unwrap() error { return e.cause }
+
+// silentExit returns an exitError and silences cobra's error output for
+// the given command. Use when the caller has already printed any user-
+// facing message and just needs the process to exit with a specific code.
+// Must only be called from a single-threaded context (RunE, post-join code),
+// not from goroutines that share cmd.
+func silentExit(cmd *cobra.Command, code int) error {
+	cmd.SilenceErrors = true
+	return &exitError{code: code}
+}
+
+// silenceIfExit silences cobra's error output for cmd when err is an
+// *exitError, then returns err unchanged. Use at top-level RunE return
+// points whose error may have originated in concurrent code that could
+// not safely mutate cmd itself.
+func silenceIfExit(cmd *cobra.Command, err error) error {
+	if _, ok := err.(*exitError); ok {
+		cmd.SilenceErrors = true
+	}
+	return err
+}
+
+// usageErr re-enables the usage block for cmd and returns err. Use inside
+// RunE for validation that's semantically part of the invocation contract
+// (mutually exclusive flags, missing required values, bad enum values).
+// PersistentPreRunE on the root silences usage for runtime errors; this
+// flips it back for the specific case of "the user invoked me wrong."
+func usageErr(cmd *cobra.Command, err error) error {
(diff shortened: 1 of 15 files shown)


cmd/roborev/helpers.go
```diff
@@ -9,17 +9,80 @@ import (
...
+// should still tell the user what went wrong. Verdict-based exits use
+// the bare *exitError sentinel; their callers silence cobra separately
+// via silenceIfExit because that path has already shown the review
+// output (or chose not to in quiet mode).
+func quietExit(cmd *cobra.Command, err error) error {
+	if err == nil {
+		return nil
+	}
+	cmd.SilenceUsage = true
+	if _, ok := err.(*exitError); ok {
```
mariusvniekerk, line 80: this should using errors.As. the casting approach is not great

cpcloud: probably need a CLAUDE.md or some such to steer it away from this behavior

mariusvniekerk: kinda surprised it survived contact with golangcilint

mariusvniekerk: can use this as is and make a follow-up to add https://golangci-lint.run/docs/linters/configuration/#errorlint

mariusvniekerk (approved):

PR #730, opened 2026-05-20
Prefer Antigravity CLI for Gemini agent

mariusvniekerk:
Due to google deprecating the gemini cli we need this one.  The antigravity CLI lacks a LOT of the functionality that we currently rely on (like --model) so we are forced to make some compromises :(

## Summary
- prefer the Antigravity `agy` CLI for the existing Gemini agent while falling back to legacy `gemini`
- invoke Antigravity in print mode, pipe prompts over stdin, and map review/agentic permissions to `--sandbox` and `--dangerously-skip-permissions`
- keep legacy Gemini stream-json handling and document the preferred Antigravity install path

## Validation
- `printf 'Return exactly: ok\\n' | agy --print --print-timeout 30s --sandbox`\n- `go test ./internal/agent`\n- `go vet ./...`\n- `go test ./...`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ REPORT.md
 .dolt/
 *.db
 .codex
+.antigravitycli/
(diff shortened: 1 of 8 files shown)


mariusvniekerk:
given that this effectively removes the ability to specify a model / reasoning effort for gemini (other than what it used last) should we make this change now?  

Its going to be forced on us regardless soonish.

PR #732, opened 2026-05-21
Add agent plugin manifests

mariusvniekerk:
## Summary
- add Codex plugin metadata pointing at the existing Codex roborev skill tree
- add Claude plugin and marketplace metadata pointing at the existing Claude roborev skill tree
- keep plugin distribution on the same skill sources used by the embedded installer

## Validation
- jq . .codex-plugin/plugin.json .claude-plugin/plugin.json .claude-plugin/marketplace.json

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,0 +1,20 @@
+{
+  "name": "roborev",
+  "description": "Marketplace for roborev agent skills.",
+  "owner": {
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
The review agent is wrong here.  the path resolution both of these used is based on the root of the repo not relative to the file

cpcloud:
This needs something to update these files when a release is cut. I'll clank on it.

PR #734, opened 2026-05-22
Write oversized diff snapshots under repo-local tmp

mariusvniekerk:
## Summary
- Write oversized diff snapshots into a repo-local directory by default instead of OS temp space
- Add `snapshot_dir` repo config so repos can choose a different local snapshot root
- Make `roborev init` ensure the snapshot directory is ignored in `.gitignore`
- Add tests for snapshot path resolution, init gitignore behavior, and snapshot file placement

## Testing
- `go test ./...`
- `go vet ./...`
- Added unit tests covering default and configured snapshot directories, traversal rejection, and init ignore handling


mariusvniekerk:
> Keep daemon-created snapshots outside the user checkout and explicitly expose that directory to agents, or ensure background jobs use an isolated worktree/scratch checkout before writing repo-local snapshots.

This is at odds with repo-level sandbox isolation so achieving that goal is impossible.

PR #737, opened 2026-05-23
feat: show agentsview cost estimate in the queue view

wesm:
## Summary

- `internal/tokens` selects the agentsview usage command via a capability probe: `agentsview session usage <id> --format json` on agentsview >= 0.30.0, falling back to the deprecated `token-use` on 0.15.0-0.29.x. The new `cost_usd`/`has_cost` fields are parsed into `Usage`.
- Usage exit codes 2 (session not found) and 3 (found, no token/cost data) are treated as "no usage" rather than errors. The legacy `token-use` exit-1 empty-output case is still handled, so older agentsview keeps working.
- `FormatSummary` appends `· ~$0.42` when a cost estimate is present (the review detail view renders this automatically); `FormatCost` returns the cost alone.
- The TUI queue gains a default-visible "Cost" column showing the per-job estimate, blank when the model is unpriced or usage has not been fetched yet.
- Cost populates once agentsview 0.30.0 (kenn-io/agentsview#536) is installed; on older versions the column stays blank and token counts are unaffected. A `backfill-tokens` run refreshes existing jobs.

### DIFF
--- a/cmd/roborev/backfill_tokens.go
+++ b/cmd/roborev/backfill_tokens.go
@@ -17,7 +17,7 @@ func backfillTokensCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "backfill-tokens",
 		Short: "Backfill token usage for completed jobs via agentsview",
-		Long: `Scan completed jobs that have a session ID but no token usage data,
+		Long: `Scan completed jobs that have a session ID but missing token cost data,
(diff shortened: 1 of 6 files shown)


wesm:
This is tested and working for me (see `roborev backfill-tokens` for populating history) with agentsview 0.30.0

PR #752, opened 2026-05-27
fix(prompt): ground reviewer version checks in project toolchain

cpcloud:
## Summary

- Add a shared review system-prompt instruction directing reviewers to verify version- and availability-related findings against the project's real toolchain and dependency manifests (go.mod, package.json + lockfiles, pyproject/requirements, etc.) for each changed file's language, instead of judging from stale model memory. This stops false positives like flagging Go's `sync.WaitGroup.Go` as nonexistent.
- Lists common manifests per language but is explicitly non-exhaustive, and handles multi-language changes by consulting every involved language's manifests.
- Keeps accuracy in both directions: a confirmed call to an API, builtin, or syntax that genuinely does not exist for the configured versions is still reported as a real bug.
- Injected at the single `GetSystemPrompt` chokepoint via a new `promptSuffix` helper, covering all agents and the review/range/dirty/security/design prompts; excluded from `run` (raw passthrough) and `address` (fix) prompts.
- Add `TestGetSystemPrompt_ToolchainVerificationInstruction` covering inclusion across agents and review types plus exclusion for `run`/`address`.

### DIFF
--- a/internal/daemon/worker_test.go
+++ b/internal/daemon/worker_test.go
@@ -825,7 +825,7 @@ func TestProcessJob_LargeDiffUsesExternalSnapshotWithoutOversizedPrompt(t *testi
 		ReviewFn: func(ctx context.Context, repoPath, commitSHA, p string, output io.Writer) (string, error) {
 			agentCalled = true
 			capturedPrompt = p
-			require.LessOrEqual(t, len(p), 4096, "submitted prompt must stay within configured cap")
+			require.LessOrEqual(t, len(p), 6000, "submitted prompt must stay within configured cap")
 			match := snapshotRE.FindStringSubmatch(p)
 			require.NotNil(t, match, "large diff prompt should reference a snapshot file")
 			snapshotPath := match[1]
@@ -840,7 +840,10 @@ func TestProcessJob_LargeDiffUsesExternalSnapshotWithoutOversizedPrompt(t *testi
 
 	tc := newWorkerTestContext(t, 1)
 	cfg := config.DefaultConfig()
-	cfg.DefaultMaxPromptSize = 4096
+	// Keep the cap above the system-prompt size so the snapshot-reference
+	// fallback still fits; the large diff below far exceeds it either way,
+	// so the external-snapshot path still triggers.
+	cfg.DefaultMaxPromptSize = 6000
 	tc.reconfigurePool(cfg)
 
 	var content strings.Builder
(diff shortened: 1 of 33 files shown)


cpcloud:
The Medium finding on `34b3433` (golden fixtures under `internal/prompt/testdata/golden` not updated) is a false positive:

- There is no `internal/prompt/testdata/golden` directory, and no `golden` / `update-golden` references anywhere in `internal/prompt/`. The suggested `go test -update-golden ./internal/prompt/` references a flag that is not defined, so it would error rather than regenerate anything.
- The prompt tests assert with `assert.Contains` on specific substrings, not full-output equality, so appending the instruction cannot cause drift. `go test ./internal/prompt/` passes.

No code change is warranted.

cpcloud:
Update: superseding my earlier false-positive comment above.

That comment was accurate for the commit the bot reviewed (`34b3433`), which was based on the previous prompt architecture — at that point there was no `internal/prompt/testdata/golden` directory and no `-update-golden` flag.

This branch has since been rebased onto current `main`, which refactored the prompt system to a template-based design **with** golden-fixture tests (`internal/prompt/golden_test.go`, `internal/prompt/testdata/golden/`). So the finding now applies, and I've handled it: regenerated the 18 affected goldens with `go test ./internal/prompt/ -update-golden`. The `address_*` goldens are intentionally unchanged, since the fix prompt is excluded from the toolchain-verification instruction.

cpcloud:
Good catch, and partly actionable. The underlying behavior is real but pre-existing: `newPromptBuildContext` hard-caps the system prefix itself (`requiredPrefix = hardCapPrompt(..., promptCap)`) and sets `bodyLimit = max(0, promptCap - len(requiredPrefix))`, so any `promptCap` below the system-prompt size truncates the prefix and zeroes the body — independent of this PR. The toolchain instruction here just raised the prefix ~715B (~3.3KB -> ~4.0KB), which tipped the `TestGoldenPrompt_SingleTruncatedDiff` golden over its artificial 4000-byte cap (the prompt was truncated mid-instruction with no `## Current Commit`/diff).

At the default cap (256KB) this is unreachable, so I haven't re-architected truncation in this PR. Instead:

- Raised that test's cap above the system-prompt size (commit `741a37a`) so the truncated prompt again includes commit metadata and a `git show` diff fallback, keeping the test meaningful.
- Filed #753 to track the real hardening (reserve a minimum body budget, make optional system guidance shrinkable under tight caps, and error rather than silently emit a truncated prefix), since that's a change to shared truncation logic beyond this PR's scope.

PR #754, opened 2026-05-27
update: use HTML redirect endpoint to avoid GitHub API rate limit

wesm:
## Summary

- `scripts/install.sh` and the in-binary `roborev update` previously called `api.github.com/repos/.../releases/latest`, which is rate-limited to 60 req/hr per IP for unauthenticated callers. Users behind shared NAT, VPN, or CGNAT hit 403s.
- Both paths now use the HTML `https://github.com/roborev-dev/roborev/releases/latest` endpoint, which 302-redirects to `/releases/tag/<tag>` and isn't subject to the same rate limit.
- `scripts/install.sh` reads the redirect target via `curl -fsSLI -w '%{url_effective}'` (or `wget --server-response`).
- `scripts/install.ps1` uses `Invoke-WebRequest -MaximumRedirection 0` and parses the Location header, with separate handling for the PowerShell 5.x `WebException` path.
- `internal/update/update.go` gains `resolveLatestTag` (follows the redirect with `CheckRedirect: ErrUseLastResponse`) and `fetchContentLength` (HEAD request for asset size). Download and `SHA256SUMS` URLs are constructed from the tag.
- Drops the now-unused `Release`/`Asset` structs, `findAssets`, `resolveAsset`, `resolveChecksum`, `fetchLatestRelease`, `fetchJSON`, and the release-body checksum fallback; every release publishes `SHA256SUMS` as an asset.
- Ports msgvault [052f42ff](https://github.com/wesm/msgvault/commit/052f42ff196260815eb73bd38a4e64668726111b).

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -22,9 +22,13 @@ import (
 )
 
 const (
-	githubAPIURL  = "https://api.github.com/repos/roborev-dev/roborev/releases/latest"
-	cacheFileName = "update_check.json"
-	cacheDuration = 1 * time.Hour
+	// githubLatestReleaseURL is the HTML endpoint that 302-redirects to
+	// /releases/tag/<tag>. Unlike api.github.com it is not rate-limited
+	// at 60 req/hr per IP for unauthenticated callers.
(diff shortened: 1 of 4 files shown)


wesm:
Hallucinated

PR #769, opened 2026-06-02
Add build-time update disable flag

cpcloud:
## Summary
- add an ldflags-controlled update.Disabled switch for package-managed builds
- make update checks return no result when self-update is disabled, suppressing TUI update prompts
- make roborev update print package-manager guidance instead of checking or installing

## Verification
- go test ./...
- go vet ./...
- golangci-lint@v2.10.1 run ./...
- go build -ldflags "-X go.kenn.io/roborev/internal/update.Disabled=true" -o /tmp/roborev-disable-update ./cmd/roborev
- /tmp/roborev-disable-update update

Refs #610

### DIFF
--- a/cmd/roborev/main_test.go
+++ b/cmd/roborev/main_test.go
@@ -27,6 +27,7 @@ import (
 	"go.kenn.io/roborev/internal/daemon"
 	"go.kenn.io/roborev/internal/git"
 	"go.kenn.io/roborev/internal/storage"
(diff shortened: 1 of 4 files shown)


mariusvniekerk (approved):

cpcloud:
Closing this for now while we move roborev self-update over to go.kenn.io/kit/selfupdate first.

PR #774, opened 2026-06-03
fix: preserve commit review when auto-design classifier fails

cpcloud:
## Summary
- add an HTTP enqueue regression covering auto-design classifier failure
- verify the primary commit review stays a normal review job when classifier setup fails
- keep the auto-design failure isolated to a skipped auto_design row

## Validation
- go fmt ./...
- go build ./...
- go test ./...
- go vet ./...
- make lint-ci

### DIFF
--- a/internal/daemon/e2e_auto_design_test.go
+++ b/internal/daemon/e2e_auto_design_test.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


cpcloud:
Drafting because I am not yet convinced the behavior is actually what I want.

cpcloud:
Closing this because the branch only adds regression coverage for behavior that already exists; it does not contain the behavioral fix implied by the PR title.

PR #787, opened 2026-06-04
Fall back to main repo for gitignored .roborev.toml in worktrees

cwensel:
## Summary

- A `.roborev.toml` that is gitignored (or otherwise untracked) lives only in the main checkout's working tree, so it is invisible from a linked worktree's directory. Reviews and fix jobs that run against a worktree path silently lost the per-repo config.
- `LoadRepoConfig` now retries at the main repository root (resolved via the git common dir, using the existing `git.GetMainRepoRoot`) when no `.roborev.toml` exists at the given path. The extra git resolution only runs on the absent path, so the common case is unchanged.
- A worktree-local `.roborev.toml` still takes precedence when present.
- Symlinked `.roborev.toml` files already work (loading follows symlinks); added a test to pin this.
- Added tests for the worktree fallback, worktree-local precedence, and symlink cases.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -520,48 +520,52 @@ func (c *Config) migrateDeprecated(md toml.MetaData) {
 	}
 }
 
-// LoadRepoConfig loads per-repo config from .roborev.toml.
+// RepoConfigPath returns the .roborev.toml path that should be read for
+// repoPath, applying the linked-worktree fallback.
 //
-// When repoPath has no .roborev.toml, it falls back to the main repository
-// root. This matters for linked worktrees: a .roborev.toml that is gitignored
-// (or otherwise untracked) lives only in the main checkout's working tree, so
-// it is invisible from a worktree's directory. The fallback resolves the main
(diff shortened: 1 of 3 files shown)


cwensel:
Addressed in 666aa29: extracted the path resolution into `RepoConfigPath`, which applies the worktree → main-checkout fallback once, and routed both `LoadRepoConfig` and `LoadRawRepo` through it so decoded and raw config loads agree on the authoritative file. Added a test (`TestLoadRawRepoWorktreeFallback`) covering explicit-key detection via the fallback.

cwensel:
Addressed in 338f69a: the fallback now only applies when the main-checkout `.roborev.toml` is untracked (the gitignored / machine-local case this targets), checked via `git ls-files`. A tracked `.roborev.toml` is branch-specific, so a worktree on a branch that lacks it no longer inherits the main checkout's copy — the worktree's own absence wins. Added `TestLoadRepoConfigWorktreeNoFallbackForTracked` to cover this.

cpcloud:
Thanks @cwensel!

PR #790, opened 2026-06-04
ci: update Nix vendorHash on Dependabot PRs

cpcloud:
## What changed

Adds a Dependabot-only workflow that repairs `flake.nix` when Go dependency PRs change `go.mod` or `go.sum` and the `buildGoModule` `vendorHash` is stale.

The workflow:

- runs only for same-repository Dependabot PRs that touch Go module files
- runs `go mod tidy` through the flake dev shell
- replaces `vendorHash` with a fake hash, runs `nix build '.#default'`, and extracts Nix's reported `got: sha256-...` value
- writes the real hash back to `flake.nix`
- verifies the repaired package with `nix build '.#default' --no-link -L`

## GitHub-script commit path

The final step uses pinned `actions/github-script` instead of `git commit && git push` so the workflow can create commits that satisfy GitHub's bot signature verification.

The script reads the changed `go.mod`, `go.sum`, and `flake.nix` files from the checked-out PR branch, then uses GitHub's Git Database API to:

- fetch the current HEAD commit and its base tree
- create a new tree containing the changed file contents
- create a commit with that tree and the PR branch HEAD as its parent
- omit custom author, committer, and signature fields so GitHub can apply bot verification
- check `nextCommit.verification.verified` before moving the branch ref
- update `heads/${HEAD_REF}` only after the created commit is verified

The job grants `contents: write`, which is the permission the Git Database API needs to create the commit and update the Dependabot branch ref.

## Notes

The workflow stays limited to same-repository Dependabot PRs. If repository or organization settings prevent Dependabot-triggered workflows from receiving a write-capable `GITHUB_TOKEN`, the workflow will fail at the GitHub API step instead of silently producing an unverified or partial branch update.

## Verification

- `nix run 'nixpkgs#actionlint' -- .github/workflows/update-vendor-hash.yml`
- `git diff --check upstream/main..HEAD -- .github/workflows/update-vendor-hash.yml`
- `rg -n 'cat <<<|printf|echo[^\n]*\|' .github/workflows/update-vendor-hash.yml`
- `nix build '.#default' --no-link -L`

### DIFF
--- a/.github/workflows/update-vendor-hash.yml
+++ b/.github/workflows/update-vendor-hash.yml
@@ -50,6 +50,7 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.ref }}
           fetch-depth: 0
+          persist-credentials: false
 
       - uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3  # v31.10.6
 
--- a/.github/workflows/update-vendor-hash.yml
+++ b/.github/workflows/update-vendor-hash.yml
@@ -1,0 +1,175 @@
+name: Update Nix Vendor Hash
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - go.mod
+      - go.sum
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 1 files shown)


wesm:
Invalid

PR #794, opened 2026-06-05
Add generated public daemon client

mariusvniekerk:
## Summary
- add a public pkg/client generated from the Huma OpenAPI contract with doordash-oss/oapi-codegen-dd
- check in the OpenAPI 3.1 YAML and split generated Go client files for reviewable diffs
- update the TUI to exercise the public client while preserving status/body based error handling

Local validation included go vet ./... and go test ./....


mariusvniekerk:
moans about generated code.  not valid.

PR #799, opened 2026-06-05
fix(release): create release before GoReleaser

cpcloud:
Reintroduces the GoReleaser release path on top of the upstream revert, but keeps release metadata outside of GoReleaser.

The release workflow now validates the pushed tag, creates the GitHub release from that tag before any artifacts are built, and then runs GoReleaser as the dependent artifact job. GoReleaser is configured to preserve existing release notes and has changelog generation disabled, so release notes stay owned by the GitHub release creation step.

This also gates the Homebrew tap update to `kenn-io/roborev`, which lets fork tag runs exercise release creation and GoReleaser artifacts without touching the shared tap.

The main tradeoff is that fork release tests still need a tag matching the existing `vMAJOR.MINOR.PATCH` validation unless we decide to loosen that separately for prerelease-style test tags.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,122 +30,61 @@ jobs:
             echo "version=${TAG_NAME#v}"
           } >> "$GITHUB_OUTPUT"
 
-  build:
+  create-release:
     needs: validate-tag
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        include:
-          - goos: linux
-            goarch: amd64
-          - goos: linux
-            goarch: arm64
-          - goos: darwin
-            goarch: amd64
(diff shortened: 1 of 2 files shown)


mariusvniekerk (approved):

cpcloud:
`keep-existing` is exactly what we want. `append` will append whatever goreleaser thinks the release notes should be.

cpcloud:
I've got a clanker going on the previous run's windows test failure that popped up

cpcloud:
Looks like it's popping up again. Unrelated, so we can merge this as is.

wesm:
rebased

cpcloud:
We can do the draft flow later if desired.

PR #800, opened 2026-06-05
Fix agent hook Windows git -C parsing

cpcloud:
Fixes a Windows-only CI failure in the agent hook commit-counting tests.

The hook command splitter treated every backslash as a shell escape, so Windows drive paths in `git -C` commands lost their separators before `commandGitDir` could resolve the target repository. That made commit tracking fall back to the hook cwd or miss the target repo entirely, which broke commit reminder attribution on Windows.

This keeps the existing lightweight command scanning approach, preserves POSIX escaped-space handling, and narrows backslash escaping so quoted and unquoted Windows `C:\...` paths remain intact. The new regression covers both Windows path forms plus the POSIX escaped-space case that the parser still needs to support.

### DIFF
--- a/internal/agenthook/state.go
+++ b/internal/agenthook/state.go
@@ -663,7 +663,13 @@ func shellFields(command string) []string {
 	expansionDepth := 0
 	inToken := false
 	pendingExpansion := false
-	for _, r := range command {
+	runes := []rune(command)
+	for i, r := range runes {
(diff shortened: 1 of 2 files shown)


wesm:
This might not be necessary after #797 was just merged

cpcloud:
Great! Closing.

PR #814, opened 2026-06-08
Add queue pause controls

mariusvniekerk:
Adds an explicit way to pause daemon queue processing without changing worker counts or restarting the daemon.

Summary:
- Adds `roborev pause` and `roborev unpause` commands backed by daemon API endpoints.
- Persists the pause state in SQLite so it survives daemon restarts.
- Makes workers skip claiming new jobs while paused, while allowing already-running jobs to finish.
- Includes `queue_paused` in daemon status JSON and shows the paused state in the TUI worker header.

Checked locally with `TMPDIR=/tmp go test ./...` and `TMPDIR=/tmp go vet ./...`.

### DIFF
--- a/cmd/roborev/daemon_lifecycle.go
+++ b/cmd/roborev/daemon_lifecycle.go
@@ -243,6 +243,15 @@ func startDaemon() error {
 		fmt.Println("Starting daemon...")
 	}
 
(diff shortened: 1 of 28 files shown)


wesm:
rebasing

PR #815, opened 2026-06-08
Fix slow TUI queue startup from repeated display-name lookups

vbdrs:
Summary:
  - Deduplicate repo paths before refreshing TUI display-name cache during job list updates.
  - Avoid repeated synchronous `.roborev.toml` lookups for many jobs from the same repo.
  - Add a focused regression test for stable repo-path deduping.

  Context:
  On live queue data, `/api/jobs` returned quickly, but the TUI spent several seconds refreshing display names once
  per job row before first paint. With many rows from the same repo, startup stayed on `Loading...` unnecessarily and would stutter while navigating the entries.
  This change refreshes each repo path once per jobs update while preserving the ability to pick up display-name
  config changes.

### DIFF
--- a/cmd/roborev/tui/display_name_test.go
+++ b/cmd/roborev/tui/display_name_test.go
@@ -1,0 +1,22 @@
+package tui
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.kenn.io/roborev/internal/storage"
+)
+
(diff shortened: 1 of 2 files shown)


wesm:
Thanks for investigating this, I'll look at this and get a bugfix 0.57.1 release out today. I fixed another perf related issue in https://github.com/kenn-io/roborev/commit/c70b075289b9c6447461da45949c44d4324c14e9

Issue #845, opened 2026-06-11
Daemon still opens visible console windows on Windows (rapid flashing) after #840


cderv:
On Windows with Git Bash, running a command that auto-starts the daemon (`roborev status`, `roborev tui`, …) makes console windows rapidly open and close. It continues until `roborev daemon stop`.

#840 lists, among its fixes:

> Bump go.kenn.io/kit to v0.1.5 so detached daemons run on a hidden console and their git/agent children stop opening visible console windows

I built from current `main` (commit `27b1277`, which includes #840) and still see the flashing. I traced it with a process-creation watch while running `roborev status`: the auto-started `roborev daemon run` process spawns a steady stream of `git.exe`, and each one gets its own `conhost.exe` / `OpenConsole.exe` — a new console window — roughly every 0.8s. The queue was empty (no jobs running), and I have 13 repos registered.

<details>
<summary>process watch (trimmed)</summary>

```
pid=49964 roborev.exe | "...\roborev.exe" status
pid=31240 ppid=49964 roborev.exe | ...\roborev.exe daemon run
pid=39832 ppid=31240 git.exe | git rev-parse --show-toplevel
pid=10080 ppid=39832 conhost.exe
pid=18908 ppid=31240 git.exe | git config --global --includes -z --get-all safe.directory
pid=30856 ppid=31240 git.exe | git rev-parse --git-common-dir
pid=53000 ppid=31240 git.exe | git rev-parse --git-dir
... (repeats ~every 0.8s; each git.exe spawns its own conhost/OpenConsole)
```
</details>

It may be coming from the daemon's git subprocess spawns rather than the paths #840 already covered — but I'm not certain. Since this is Windows-specific and you may not have a Windows machine handy, I'm happy to dig in and test patches or instrumented builds on my environment to narrow it down.

Environment: Windows 11, Git Bash, roborev v0.57.1 (also reproduced with a build from `main` at `27b1277`).

Related: #834, #840

cderv:
I dug into this a bit more and I think I've found why #840 didn't take effect on Windows — though please correct me if I'm misreading it.

#840 bumped `go.kenn.io/kit` to v0.1.5, which runs detached children on a hidden console (`CREATE_NO_WINDOW`) instead of `DETACHED_PROCESS`, so descendants like `git.exe` don't allocate their own visible console. The unix daemon launcher goes through that path:

[...]

PR #846, opened 2026-06-11
Launch the Windows daemon on a hidden console to stop console flashing

cderv:
## Summary

- On Windows the detached-daemon launcher used `DETACHED_PROCESS`, which leaves the daemon with no console at all. Every console-subsystem child it spawns (the steady stream of `git.exe`, agent CLIs) then allocates its own visible console window, so a backgrounded daemon flashes terminal windows roughly once a second until it is stopped (#845).
- #840 bumped `go.kenn.io/kit` to v0.1.5 specifically to fix this — kit's detached launcher runs the child on a hidden console (`CREATE_NO_WINDOW`) that descendants inherit — and wired the unix launcher to it, but the Windows launcher kept its hand-rolled `DETACHED_PROCESS` path. The two flags are mutually exclusive (`CREATE_NO_WINDOW` is ignored when `DETACHED_PROCESS` is set), so the kit fix never took effect on Windows.
- Route both platforms through kit's `StartDetached`, which already covers every option the Windows launcher used, so the daemon runs on a hidden console on Windows too and its git/agent children no longer open visible windows. The individual git call sites need no change.
- Add a Windows regression test asserting the detached daemon uses `CREATE_NO_WINDOW` and not `DETACHED_PROCESS`.


wesm:
I'm going to try to cut a roborev release this morning to address some of the Windows issues, let me know when you think this is mergeable

cderv:
I am still trying to solve this the right way. I made it a draft because I want to test first locally

cderv:
@wesm — I'd hold this one back from the release; it's not mergeable. I tested the build locally and it trades the flashing for a worse regression: the daemon no longer starts.

`roborev daemon start` fails with `failed to start daemon: context deadline exceeded`; `roborev daemon run` foreground is fine, so the daemon's healthy and it's the detached launch that breaks. The detached process spawns and stays alive but never binds, never publishes its runtime file, and never writes its first log line.

delve on the hung `daemon run`, main goroutine:

```
lipgloss/v2.BackgroundColor
  -> queryBackgroundColor -> queryTerminal
  -> muesli/cancelreader.Read -> os.File.Read
  -> internal/poll.(*FD).readConsole -> syscall.ReadConsole   [blocks forever]
```

From what I can tell: kit's `StartDetached` gives the daemon a hidden console (`CREATE_NO_WINDOW`) that still has a real input buffer and reports as a TTY, so lipgloss's startup background-color query (via the output styling) blocks on `ReadConsole` waiting for a reply that never comes. The old `DETACHED_PROCESS` path has no console, so the query is skipped and the daemon starts (it just flashes). Foreground works because the real terminal answers. If that's right, kit's hidden console (kenn-io/kit#14) isn't safe for the daemon while it does lipgloss terminal detection at startup — the e2e test there spawns a short-lived child that writes a file and wouldn't hit it.

So this PR is a dead end. The real fix is probably to keep the daemon detached (no console — works today) and suppress the console windows on the git/agent children instead — same idea as the existing `HiddenCommand` wrapper, extended to the git spawns — or stop the daemon doing terminal detection at all. Both are more involved than this one-liner, so I doubt it makes this morning's release. I'll keep digging; more detail in #845.

cderv:
Opening #848 with the child-stamping approach from my last comment on #845 — keep the daemon on `DETACHED_PROCESS` and stamp `CREATE_NO_WINDOW` on each child roborev spawns directly (`internal/git`, agents, worktree, hooks, TUI tasklist poll).

The startup git burst from kit's Runner (each invocation spawning unguarded `git config safe.directory` calls) is filed separately as kenn-io/kit#18; that and #848 together close #845.

Closing this draft.

PR #850, opened 2026-06-12
Make update check survive GitHub API rate limits

wesm:
## Summary

- The watchdog caught `roborev update` failing on GitHub-hosted runners: the release check calls `api.github.com` unauthenticated, and shared runner egress IPs routinely exhaust the per-IP rate limit, returning 403.
- When the API check fails, fall back to resolving the latest release through github.com: follow the `/releases/latest` redirect to the tag (handles repo renames) and read the asset checksum from `SHA256SUMS`. Web and download endpoints are not subject to API rate limits and need no token. If both paths fail, the error reports both causes.
- Honor `GH_TOKEN`/`GITHUB_TOKEN` for the API check when set, scoped to the API host only — asset downloads redirect to pre-signed storage URLs that reject `Authorization` headers.
- Drop the duplicate "check for updates:" error wrap in the CLI; kit already prefixes it (the watchdog log showed `check for updates: check for updates: GitHub API returned 403 Forbidden`).

### DIFF
--- a/cmd/roborev/update.go
+++ b/cmd/roborev/update.go
@@ -365,9 +365,10 @@ launchd or systemd).`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Println("Checking for updates...")
 
+			// kit already wraps check errors with "check for updates:".
 			info, err := update.CheckForUpdate(true) // Force check, ignore cache
 			if err != nil {
-				return fmt.Errorf("check for updates: %w", err)
+				return err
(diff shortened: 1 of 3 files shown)


mariusvniekerk (approved):

mariusvniekerk:
We may want to serve a static bit of json on the docs page that has the most recent version on it at some point to avoid the issue.

wesm:
Good point. I can clank on that

PR #853, opened 2026-06-17
Tighten security review prompt against noisy findings

mariusvniekerk:
Security review output has been drifting toward hardening advice that does not change attacker capability, especially environment-variable findings that assume same-user local access is a meaningful boundary. That noise makes reviews less actionable and would pollute an eval bench of past reviews by rewarding security theater instead of concrete changes in risk.

This draft tightens both the commit-review and analyze-security prompts around an exploitability burden of proof: a finding needs a realistic actor, a real trust boundary or security control, and a concrete asset or privilege at risk. It still leaves room for material weakening of security controls without requiring a turnkey exploit, but explicitly filters local-only env-var hardening, same-user process access, generic dependency pinning concerns, and low-severity defense-in-depth findings.

The prompt wording is locked in with direct prompt assertions and the security-review golden snapshot, so follow-up eval work can compare old and new agent behavior against known noisy historical reviews. Local validation run: `go fmt ./...`, `go vet ./...`, `go test ./...`.

### DIFF
--- a/internal/prompt/analyze/analyze_test.go
+++ b/internal/prompt/analyze/analyze_test.go
@@ -67,6 +67,18 @@ func TestSecurityPromptRequiresActionableFindings(t *testing.T) {
 	assert.Contains(t, prompt, "concrete", "prompt should discourage generic findings")
 }
 
+func TestSecurityPromptDiscouragesSecurityTheater(t *testing.T) {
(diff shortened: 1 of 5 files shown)


wesm:
LGTM. Seems like this will improve the quality of security reviews

PR #854, opened 2026-06-17
Fix check-agents command to use resolved command for available agents

EwanDawson:
This PR fixes a bug in the `check-agents` health command where it checks agent availability using `agent.IsAvailable(name)` (which successfully resolves command candidates like `agy` for `gemini`), but then executes the smoke-test review by retrieving the unmodified default agent using `agent.Get(name)` (which still uses `gemini`). This causes `check-agents` to fail on systems where only `agy` is installed on the `PATH`.

With this change, `check-agents` retrieves the resolved agent via `agent.GetAvailable(name)`, which applies the resolved command candidate at runtime, ensuring health checks for `gemini` succeed when `agy` is available on the `PATH`.

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/cmd/roborev/check_agents.go
+++ b/cmd/roborev/check_agents.go
@@ -61,7 +61,20 @@ Examples:
 					continue
 				}
 
-				a, _ := agent.Get(name)
+				if !agent.IsAvailable(name) {
+					a, _ := agent.Get(name)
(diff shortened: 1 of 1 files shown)


wesm:
LGTM

PR #856, opened 2026-06-17
Restrict roborev fix discovery to failing reviews

mariusvniekerk:
Passed reviews can remain open long enough for the agent hook to prompt the bundled fix workflow, but they are not actionable work. When discovery treats every open row as fixable, the hook appears to be asking agents to fix clean reviews and the skill wastes a cycle fetching and skipping no-op jobs.

This changes discovery to exclude known pass verdicts while leaving legacy or empty-verdict rows available for the existing fetch-time fallback. It also updates the bundled fix-skill instructions so agents look for open failing reviews, not generic open review rows.

Validated with `go vet ./...`, `go test ./cmd/roborev ./internal/agenthook -count=1`, and `go test ./...`.

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -770,12 +770,23 @@ func queryOpenJobs(
 			return nil, fmt.Errorf("decode response: %w", err)
 		}
 
-		return jobsResp.Jobs, nil
+		return filterFixCandidateJobs(jobsResp.Jobs), nil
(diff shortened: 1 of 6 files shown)


mariusvniekerk:
> so discovery may still treat nil-verdict rows and non-review rows as actionable.

those rows are WEIRD and probably SHOULD demand action....

PR #858, opened 2026-06-18
Align branchless fix discovery with hook lineage

mariusvniekerk:
The stop hook can correctly decide that branchless failed review rows are actionable for the current checkout, especially when reviews were created before a branch label existed or while work was detached. After #856, `roborev fix --list` only shows failed actionable review jobs, but its branch filter still hid branchless rows that the hook was asking the agent to fix. That made the prompt look stale even when legitimate fixable review work existed.

This keeps explicit `--branch` filtering exact, while the default current-checkout path now includes branchless repo-scoped/dirty rows and concrete branchless refs that belong to the current branch lineage. Concrete refs already reachable from trunk are excluded on feature branches, so old base-branch reviews do not follow every descendant branch.

```mermaid
gitGraph LR:
  commit id: "trunk base"
  commit id: "trunk-only review excluded"
  branch feature
  checkout feature
  commit id: "A branchless review included"
  commit id: "B branchless review included"
  commit id: "C branchless review included"
  commit id: "D branch created here"
```

In the default `roborev fix` path, branchless reviews for A, B, C, and D are included because they are reachable from the current feature `HEAD` and are not already reachable from trunk. The trunk-only row is excluded on the feature branch. Passing `--branch` still requires the stored job branch to match exactly, and detached HEAD keeps using the detached-ref matching path.

Checked locally with `go test ./...` and `go vet ./...`.

### DIFF
--- a/cmd/roborev/fix.go
+++ b/cmd/roborev/fix.go
@@ -668,6 +668,15 @@ func filterReachableJobs(
 		matchBranch = gitrepo.CurrentBranch(ctx, worktreeRoot)
 	}
 	allowBranchlessLineage := branchOverride == "" && matchBranch != ""
+	var lineageMatcher *git.BranchLineageMatcher
+	lineageMatcherLoaded := false
+	lineageMatches := func(ref string) bool {
+		if !lineageMatcherLoaded {
+			lineageMatcherLoaded = true
+			lineageMatcher, _ = git.NewBranchLineageMatcherCtx(ctx, worktreeRoot, matchBranch, "HEAD")
+		}
+		return lineageMatcher != nil && lineageMatcher.Matches(ref)
+	}
 	var detachedRefs map[string]struct{}
 	if matchBranch == "" {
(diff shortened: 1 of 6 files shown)


mariusvniekerk:
Reviewed the roborev finding about using configured branch bases for branchless lineage. I do not think we should apply that recommendation as written. Branch review base selection controls what a review compares, but branchless lineage should answer whether the ref is in the current checkout ancestry relative to trunk/default.

The intended boundary is the merge-base of the repository default branch and `HEAD`, i.e. commits in `merge-base(default, HEAD)..HEAD`. That means a feature branch cut from `release/1.2` may include release-only commits after `main` forked; that is expected if those commits are in the current checkout ancestry. Using `branch.<name>.base` or a previous `--base` value would make lineage depend on review-range config and could hide branchless reviews that are genuinely in the current ancestry.

So I am not changing code for that finding as written. The current `rev-list HEAD --not <default>` implementation matches this behavior for normal histories, though we can clarify comments/tests around the merge-base framing separately.

PR #865, opened 2026-06-20
Integrate docs into roborev repository

wesm:
Roborev docs currently depend on a separate repository that is planned for deletion, while screenshots and other image media would bloat normal clones if imported directly into main. This moves the Zensical docs source into the roborev repo while keeping media in two orphan asset branches: docs-assets for curated static media and docs-generated-assets for regenerated CLI/TUI screenshots.

The branch also adds the maintainer workflow needed to operate this setup: hydration from asset refs before builds, scripts for updating each asset branch without switching the current worktree, and the kata-style manual deploy path where scripts/update-docs.sh refreshes generated assets, checks the site, and then calls make docs-deploy. The direct make docs-deploy target remains available for the final Vercel production deploy step.

The important follow-on consequence is that Vercel builds from main need the docs-assets and docs-generated-assets refs to exist on origin; both branches have been pushed. I validated the branch with make docs-check and go test ./..., and also confirmed no docs image media is tracked in main.


wesm:
This is close enough. Merging

PR #873, opened 2026-06-22
Keep status off slow repo config probes

### LINKED ISSUE #871: Windows: roborev status reports "Daemon: not running" for a healthy daemon — /api/status times out against the 2s client deadline
Since a recent update, `roborev status` on Windows reports the daemon as not running even though it's alive and serving every other endpoint. There's been a run of Windows regressions around the recent daemon/console changes (#834, #840, #845, #851); this looks like a distinct cause, so I'm filing it separately to track.

### What I see

```
roborev daemon start   -> Daemon started        (daemon is up: bound 127.0.0.1:7373, runtime file written, version matches)
roborev status         -> Daemon: not running   (false)
roborev daemon stop    -> Daemon stopped         (works)
```

This is different from #834 (an unresponsive daemon where `/api/enqueue` also timed out, fixed by #840). Here the daemon is healthy — only `status` mislabels it.

### What I found

Timing each endpoint while the daemon runs:

| Endpoint | Latency |
|---|---|
| `/api/ping` | instant |
| `/api/health` | 0.004s |
| `/api/jobs?limit=1` | 0.019s |
| `/api/status` | ~8.2s (three consecutive calls: 8.4 / 8.0 / 7.7s) |

`daemon stop` uses `/api/shutdown` and `daemon start` uses `/api/ping` (both fast), so only `status` hits the slow path.

A debugger breakpoint on process spawn showed every `/api/status` fanning out into a burst of `git` subprocesses:

```
humaGetStatus                           internal/daemon/server.go:1368
 -> autoDesignStatusForResponse         internal/daemon/server_auto_design.go
 -> config.ResolveAutoDesignHookEnabled internal/config/autodesign.go:131
 -> config.LoadRepoConfig               internal/config/config.go:601
 -> config.RepoConfigPath               internal/config/config.go:573
 -> git.GetMainRepoRoot                 internal/git/git.go:734  -> exec git
```

From what I can tell reading the code (please correct me if I'm off): `autoDesignStatusForResponse` loops every registered repo and resolves the repo config twice per repo on every `/api/status`:

https://github.com/kenn-io/roborev/blob/bd266d983297c65f0a765268ff3fa6cf028c8227/internal/daemon/server_auto_design.go#L84-L99

For a repo with no local `.roborev.toml`, `RepoConfigPath` (the worktree fallback added in #787) calls `git.GetMainRepoRoot`, which shells out to `git rev-parse`. Repos that do have a local `.roborev.toml` take the `os.Stat` path and skip git:

https://github.com/kenn-io/roborev/blob/bd266d983297c65f0a765268ff3fa6cf028c8227/internal/config/config.go#L573

On this machine: 14 registered repos, 12 without a local `.roborev.toml` — roughly 48 `git` spawns per `/api/status`, scaling linearly with repo count. I couldn't find any database involvement; the DB side is fast.

One thing I noticed: the per-repo git loop runs in the `if !enabled` branch — when auto-design is off (the default), it scans every repo to discover whether any opts in, so the cost lands on the default-off configuration.

On the message itself: `ensureDaemon()` hits `/api/ping`, succeeds, version matches — daemon confirmed alive — and then the `/api/status` request runs with a 2s timeout, times out, and the command prints "Daemon: not running", so the timeout reads as the daemon being down:

https://github.com/kenn-io/roborev/blob/bd266d983297c65f0a765268ff3fa6cf028c8227/cmd/roborev/status.go#L38-L61

### Where I tested

I've only reproduced this on Windows, where ~48 git spawns are enough to push `/api/status` past the 2s timeout. I haven't measured the same path on Linux/macOS, so I don't know whether it reaches the timeout there.

### Repro (doesn't need my setup)

1. Register 10+ repos with no local `.roborev.toml`, auto-design disabled (default).
2. Daemon running. Time `GET /api/status` (or run `roborev status`).
3. On Windows, latency scales ~linearly with repo count; once it crosses 2s, `status` prints "Daemon: not running".

Related: #787 (worktree config fallback), #661 and #852 (auto-design status routing) — the combination seems to be what put per-repo git onto the status path. #851 touches this same `internal/git` spawn path, but it's about console suppression rather than spawn count, so I don't think it changes this. I'm on Windows and happy to test a fix.

### Environment

roborev `v0.57.1-42-gbd266d9` (built from `main`), Windows 11, kit v0.1.7.

wesm:
Windows users can have a healthy daemon while `roborev status` reports it as not running because the status endpoint currently performs per-repo auto-design config discovery. In the default-off configuration, registered repos without a local `.roborev.toml` can fall into the linked-worktree config fallback and spawn `git`; with enough repos on Windows, that can exceed the CLI status deadline.

This keeps ordinary checkout config misses on a filesystem-only path while preserving the worktree fallback for actual linked worktrees, and it resolves both auto-design flags from a single loaded repo config. The CLI also now distinguishes a reachable daemon with unavailable status details from an actually missing daemon, so a slow status endpoint no longer tells users to start a daemon that is already alive.

The main reviewer consequence is that `/api/status` should stay cheap for default installations with many registered repos, while still surfacing auto-design metrics when global or per-repo config enables them. Full Go tests pass locally.

Fixes #871

### DIFF
--- a/cmd/roborev/status.go
+++ b/cmd/roborev/status.go
@@ -20,6 +20,7 @@ type statusJSONResult struct {
 	Daemon  *storage.DaemonStatus `json:"daemon,omitempty"`
 	Health  *storage.HealthStatus `json:"health,omitempty"`
 	Jobs    []storage.ReviewJob   `json:"jobs,omitempty"`
+	Error   string                `json:"error,omitempty"`
 }
 
 func statusCmd() *cobra.Command {
@@ -33,6 +34,17 @@ func statusCmd() *cobra.Command {
 				enc := json.NewEncoder(os.Stdout)
 				enc.SetIndent("", "  ")
 				return enc.Encode(result)
+			}
+			writeStatusUnavailable := func(err error) error {
+				if jsonOutput {
+					return writeJSONResult(statusJSONResult{
+						Running: true,
+						Error:   err.Error(),
+					})
+				}
+				fmt.Println("Daemon: running")
+				fmt.Printf("Status: unavailable: %v\n", err)
+				return nil
 			}
 
 			// Ensure daemon is running (and restart if version mismatch)
@@ -51,13 +63,7 @@ func statusCmd() *cobra.Command {
 			client := ep.HTTPClient(2 * time.Second)
 			resp, err := client.Get(addr + "/api/status")
 			if err != nil {
-				if jsonOutput {
-					return writeJSONResult(statusJSONResult{Running: false})
-				}
-				fmt.Println("Daemon: not running")
-				fmt.Println()
-				fmt.Println("Start with: roborev daemon start")
-				return nil
+				return writeStatusUnavailable(err)
 			}
 			defer resp.Body.Close()
 
--- a/cmd/roborev/status_test.go
+++ b/cmd/roborev/status_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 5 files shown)


wesm:
Windows verification context for follow-up agent:

The most useful check is an isolated local binary run with a temp `ROBOREV_DATA_DIR`, so it does not touch the normal daemon DB.

```powershell
gh pr checkout 873
go test ./...

New-Item -ItemType Directory -Force .tmp\rr-data, .tmp\repos | Out-Null
$env:ROBOREV_DATA_DIR = (Resolve-Path .tmp\rr-data).Path
go build -o .tmp\roborev-871.exe .\cmd\roborev

.\.tmp\roborev-871.exe daemon start
```

Then either use the existing repro shape, or create 10-20 temp repos with no `.roborev.toml`, run `.\.tmp\roborev-871.exe init --no-daemon` inside each, and time:

```powershell
Measure-Command { .\.tmp\roborev-871.exe status | Out-Null }
.\.tmp\roborev-871.exe status
.\.tmp\roborev-871.exe status --json
```

Expected: no `Daemon: not running`; status should stay comfortably under the 2s CLI deadline. For a direct before/after comparison, run the same temp `ROBOREV_DATA_DIR` against `main` and this PR branch.

wesm:
Windows validation after rebasing on current `main`:

- `go test ./...` passes locally on Windows.
- Built a temporary PR binary with an isolated `ROBOREV_DATA_DIR`.
- Started a temp daemon, registered 20 throwaway git repos with no repo `.roborev.toml`, and ran the status checks from the earlier repro note.
- `roborev status` completed in about 1.2s and reported the daemon as running.
- `roborev status --json` completed in about 62ms and reported `running: true`.

This covers the slow repo-config probe shape from #871 without touching the normal local daemon data.

mariusvniekerk (approved):

PR #874, opened 2026-06-22
Avoid terminal probe import at CLI startup

### LINKED ISSUE #872: Windows: every roborev CLI command hangs at startup under a console that does not answer the OSC-11 color query
On Windows, every roborev command — including `roborev version` — hangs at startup when stdin is a console that never answers the OSC-11 background-color query. I reproduced it in Git Bash's MSYS pty; I haven't tested other terminals. The hang is silent — no output, and the process never returns — so it reads as the binary doing nothing rather than blocking. The detached daemon is unaffected, so reviews still work; it's the foreground CLI that's unusable in this terminal. Filing alongside the other recent Windows issues to track separately.

This looks related to #683 (closed), which dealt with the same import-time terminal-probe mechanism — but that was cosmetic escape-byte leakage under tmux on Linux/macOS. This is a hard hang on Windows, so it seems like a different surface.

### What I found

Stack of a hung `roborev version` (from a debugger):

```
charm.land/lipgloss/v2/compat.init()   <autogenerated>:1   (runs at import time)
 -> lipgloss/v2.HasDarkBackground()     query.go:87
 -> lipgloss/v2.BackgroundColor()       query.go:53
 -> lipgloss/v2.queryBackgroundColor()  query.go:20
 -> lipgloss/v2.queryTerminal()         terminal.go:101
 -> muesli/cancelreader.(*fallbackCancelReader).Read   cancelreader.go:53
 -> os.File.Read -> internal/poll.readConsole -> syscall.ReadConsole   [blocks forever]
```

From what I can tell (hypothesis — please correct me if I'm misreading): `lipgloss/v2/compat` runs `HasDarkBackground()` from its package `init()`, so the probe fires at import time on every command, TUI or not. It's pulled in transitively here, which is linked into every command:

https://github.com/kenn-io/roborev/blob/bd266d983297c65f0a765268ff3fa6cf028c8227/internal/streamfmt/streamfmt.go#L16

https://github.com/kenn-io/roborev/blob/bd266d983297c65f0a765268ff3fa6cf028c8227/cmd/roborev/tui/colors.go#L5

`queryTerminal` sets a 2s timeout that calls `rd.Cancel()`, but on Windows the cancelreader falls back to muesli's `fallbackCancelReader`, whose `Cancel()` is a no-op against a blocked `ReadConsole` — so the timeout doesn't interrupt the read and it blocks.

The detached daemon escapes it: with `DETACHED_PROCESS` it has no console, so the query returns immediately. An isolated foreground run in a normal console, and `daemon start` / `daemon run`, all work fine — so the trigger seems to be a console that's present but doesn't answer the query.

### Related

This is the same `ReadConsole`-blocks-forever probe documented in kenn-io/kit#19, which was fixed for the daemon launcher by keeping it on `DETACHED_PROCESS` (kenn-io/kit#20). That fix covers the daemon process; CLI commands reach the probe through the import-time `compat.init()` path instead, which the daemon-side change doesn't touch. I'm on Windows and happy to test a change.

### Environment

roborev `v0.57.1-42-gbd266d9` (built from `main`), Windows 11, kit v0.1.7. Reproduced in Git Bash (MSYS pty).

wesm:
Fixes #872.

Windows users can hit a foreground CLI startup hang when their console exposes stdin but never answers the OSC-11 background-color query. The failure happens before command dispatch because `lipgloss/v2/compat` performs that query from package init, so even commands like `roborev version` can appear to do nothing.

This removes the impure compat package from the roborev command dependency graph and keeps adaptive colors behind a local helper that has no terminal I/O at init. Auto color selection still uses the existing stream formatter detection point once execution reaches code that is intentionally rendering terminal output.

The important review boundary is startup behavior: the binary should not regain an import-time terminal probe while registering commands. A dependency-graph regression test now guards that directly.

Windows validation also exposed script-test harness assumptions around CRLF shell scripts and raw Windows paths passed into Bash. The PR now normalizes shipped shell scripts to LF and makes the script tests hand Bash-visible paths and remotes to WSL/Git Bash.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
+*.sh text eol=lf
 *.gotmpl text eol=lf
 *.golden text eol=lf
--- a/cmd/roborev/startup_imports_test.go
+++ b/cmd/roborev/startup_imports_test.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRoborevStartupAvoidsTerminalProbeImports(t *testing.T) {
+	cmd := exec.Command("go", "list", "-deps", ".")
+	out, err := cmd.CombinedOutput()
+	require.NoError(t, err, "go list -deps failed:\n%s", out)
+
+	deps := strings.Fields(string(out))
+	assert.NotContains(t, deps, "charm.land/lipgloss/v2/compat")
+}
--- a/cmd/roborev/tui/colors.go
+++ b/cmd/roborev/tui/colors.go
@@ -1,13 +1,11 @@
 package tui
 
 import (
-	"charm.land/lipgloss/v2"
-	"charm.land/lipgloss/v2/compat"
+	"image/color"
+
+	"go.kenn.io/roborev/internal/termstyle"
 )
 
-func adaptiveColor(light, dark string) compat.AdaptiveColor {
-	return compat.AdaptiveColor{
-		Light: lipgloss.Color(light),
-		Dark:  lipgloss.Color(dark),
-	}
+func adaptiveColor(light, dark string) color.Color {
+	return termstyle.AdaptiveColor(light, dark)
 }
--- a/flake.nix
+++ b/flake.nix
@@ -20,7 +20,7 @@
 
             src = ./.;
 
-            vendorHash = "sha256-rrjsevN+7xIOWZOeEFchOdfoaqB3usKXCELsgC8De2s=";
+            vendorHash = "sha256-XxmzuVbNDl3ed10JCqwkJPizrf+qzjW5OK+yiDs4Bdw=";
 
             subPackages = [ "cmd/roborev" ];
 
--- a/internal/streamfmt/streamfmt.go
+++ b/internal/streamfmt/streamfmt.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"image/color"
 	"io"
 	"os"
 	"strings"
@@ -13,9 +14,10 @@ import (
 	gansi "charm.land/glamour/v2/ansi"
 	"charm.land/glamour/v2/styles"
 	"charm.land/lipgloss/v2"
-	"charm.land/lipgloss/v2/compat"
 	"github.com/muesli/termenv"
 	"golang.org/x/term"
+
+	"go.kenn.io/roborev/internal/termstyle"
 )
 
 // Styles for TTY-mode stream output.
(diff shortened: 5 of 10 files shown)


wesm:
Helpful Windows ARM64 vetting steps for this fix:

1. Build the PR locally without installing over the normal `roborev` binary:

```powershell
git fetch origin pull/874/head:pr-874
git switch pr-874
go build -o .\roborev-issue872-arm64.exe .\cmd\roborev
```

2. Run these from Git Bash/MSYS pty on the Windows ARM64 machine:

```bash
./roborev-issue872-arm64.exe version
./roborev-issue872-arm64.exe --help
./roborev-issue872-arm64.exe config list
```

Expected: each returns promptly with no silent hang.

3. If convenient, run the same binary from PowerShell, `cmd.exe`, and Windows Terminal:

```powershell
.\roborev-issue872-arm64.exe version
.\roborev-issue872-arm64.exe --help
```

4. Check the color-mode paths do not regress startup:

```powershell
$env:ROBOREV_COLOR_MODE="auto"; .\roborev-issue872-arm64.exe version
$env:ROBOREV_COLOR_MODE="dark"; .\roborev-issue872-arm64.exe version
$env:ROBOREV_COLOR_MODE="light"; .\roborev-issue872-arm64.exe version
$env:ROBOREV_COLOR_MODE="none"; .\roborev-issue872-arm64.exe version
```

5. Best confirmation, if the old binary is available in the same Git Bash session:

- old binary hangs on `roborev version`
- PR binary returns immediately

Useful report-back details:

- Windows version
- terminal used, especially Git Bash/MSYS version if known
- ARM64 Go version from `go version`
- whether the old binary reproduced the hang
- whether the PR binary returned promptly
- any command that still hangs

The highest-signal result is: on Windows ARM64 Git Bash/MSYS pty, `roborev version` no longer hangs.

mariusvniekerk:
## Windows AMD64 verification

Verified the PR on native Windows AMD64. The startup commands returned promptly; no silent hang observed.

<details>
<summary>Environment and command details</summary>

PR checkout:

- Branch: `pr-874`
- Commit: `3fc2d87`

Host/toolchain:

- Windows: Windows 11 `10.0.26200`
- Platform: `windows/amd64`
- Go: `go version go1.26.3 windows/amd64`
- Go installed via native Windows `mise`
- mise: `2026.5.6 windows-x64`
- Git: `git version 2.54.0.windows.1`

Build/test:

```powershell
go build -o .\roborev-issue872-windows-amd64.exe .\cmd\roborev
go test ./cmd/roborev -run TestRoborevStartupAvoidsTerminalProbeImports -count=1
```

Result:

```text
ok go.kenn.io/roborev/cmd/roborev 9.509s
```

Startup checks in PowerShell:

```powershell
.\roborev-issue872-windows-amd64.exe version
.\roborev-issue872-windows-amd64.exe --help
.\roborev-issue872-windows-amd64.exe config list
```

Startup checks in `cmd.exe`:

```cmd
cmd /c .\roborev-issue872-windows-amd64.exe version
cmd /c .\roborev-issue872-windows-amd64.exe --help
```

Color mode checks in PowerShell:

```powershell
$env:ROBOREV_COLOR_MODE='auto'; .\roborev-issue872-windows-amd64.exe version
$env:ROBOREV_COLOR_MODE='dark'; .\roborev-issue872-windows-amd64.exe version
$env:ROBOREV_COLOR_MODE='light'; .\roborev-issue872-windows-amd64.exe version
$env:ROBOREV_COLOR_MODE='none'; .\roborev-issue872-windows-amd64.exe version
```

All of the above returned promptly. `version` printed `roborev dev`.

I checked for native Git Bash/MSYS coverage, but the available `bash.exe` was WSL/Linux (`x86_64-pc-linux-gnu`), so I did not count it as Windows terminal coverage.

The disposable test binary was removed afterward.

</details>

PR #896, opened 2026-06-25
Reject dirty submodule edits during refine

### LINKED ISSUE #888: Cannot commit in git submodules
My repo uses a git submodule, and Roborev doesn't seem to be able to make changes in those.

Cause when it creates changes in the submodule and tries to commit them, roborev exists with these as last log lines:
```bash
Agent completed [0:56]
Error: failed to commit changes: git commit: exit status 1: 
```

I've also had this error in the past, it seems that the subshell used for the submodule does not inherit all the git config settings?
```sh
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

wesm:
Refine runs agents in a parent repository worktree, but git submodules are separate repositories. When an agent modifies files inside a submodule, the parent repo cannot commit those file edits; it can only commit a gitlink after the submodule has its own commit. Trying to make the parent refine flow repair or commit nested submodule state is risky and unpredictable.

This PR treats submodules as a repository boundary for refine. Agents are prompted not to modify submodule contents, and refine now snapshots a small amount of submodule state so it can refuse dirty submodules, gitlink drift, or initialized submodule worktree changes before applying patches or accepting commits. If a successful hook stages a submodule gitlink change, refine rolls back the parent commit and reports the submodule path.

To keep the behavior predictable, refine does not try to clean or restore submodule worktrees. In repositories with submodules, hook-fix retry agents are not run from the parent checkout; users should fix hook failures manually or run refine from inside the submodule when the submodule itself needs changes.

Closes #888


pimlie:
Thanks for the quick PR, but note that _That keeps submodules usable as read/build context while avoiding nested submodule commit orchestration in the parent refine flow._ might not be the proper flow long term.

I'm aware the potential complexity that submodules bring, but in my case the submodule is an integral part of the actual codebase. To explain by example, the submodule I use contains the OpenAPI spec of my app. And my app generates its API services based on that OpenAPI spec.

So when I create a new feature that requires a new endpoint, I first add the endpoint to the OpenAPI spec, then generate the new Go code in my main repo, and then add the new services. Things that roborev tried to fix in the past (and I did manually based on it's logs / instructions for now), are stuff like consistent pagination markers, proper/consistent response formatting etc. Because of the way this works, these changes cannot 'just' be done in either the main repo or the openapi submodule, the logic is just too intertwined.
It would be great if roborev would be smart enough to understand this, or at least maybe make it easier? Like if the commit orchestration is too difficult, then if roborev at least write the changes in the submodule and give me instructions to commit the changes in the submodule manually then that would be awesome too :)

wesm:
I understand what you are saying, but the problem here is that `roborev refine` is not an agent harness and so dealing with the complexity of editing submodules is a big challenge. 

We actually have periodically discussed deprecating and removing the `refine` feature altogether in favor of agent skills that put the harness (e.g. codex or claude code) in the driver's seat (the agents can manage the submodules, and they have more context about how your project works), rather than a deterministic state machine. It was a nice idea, but I don't use it anymore myself and it has been a honeypot for new users who then churn off the project when it doesn't work right. My advice to new users is "don't try using refine until you have run 1000 reviews as part of your interactive agent loops".

pimlie:
> My advice to new users

Thanks. The reason why I like  to use `refine` is because I have tons of ADR's in my repo that describe how the code should work. Creating & managing these ADR's is an ever ongoing task. And `refine` might be the closest thing to ensure that PR review by an Agent is as deterministic as possible. (It isn't though, as Copilot on GH often still finds issues even after running refine)

I have had PR's where `refine` ran for 8+ hours and made 50+ commits. And when I manually review those commits, almost all make sense. The ones that don't are also often just gaps or un-clarities in my ADR's. Also, most of the commits done by refine in my repo are often about 'correctness' issues. Like not strictly necessary, just making sure all code & functionality behave similar and follow the ADRs to a T (sometimes even too much to a T, then I need to update the ADR).

Another thing I like about refine is that its automated. Like I dont wanna have to babysit an Agent while reviewing & fixing the code and I def. do not want to have to [copy/paste stuff manually](https://www.roborev.io/quickstart/#copy-and-paste-into-your-agent). Just let it rip, so I can work on something else and when `refine` is finished I can spent 1 to N hours continuously reviewing what it did all at once and revert if needed 🙏

wesm:
Makes sense, good feedback. 

I think getting refine to be submodule friendly is a bigger project — I'm open to it, especially if there are motivated folks who can help harden it and provide feedback. In the meantime, this PR seems like a good harm reduction measure: it doesn't cripple refine on projects with submodules, but it tries to avoid bad situations that can occur currently.

wesm:
This has become a mess. I'm rolling back to a simpler implementation

wesm:
Good enough here. Merging

PR #897, opened 2026-06-25
Add Codex config passthrough for -c overrides

zeroshade:
## Why

roborev passes `--ignore-user-config` to Codex so review jobs don't load the user's `~/.codex/config.toml` (skills, etc.). A side effect is that a custom `model_provider` / `[model_providers.*]` defined there is also ignored, so there was no way to point roborev's Codex at a non-default provider.

## What

Adds an `[agent.codex.config]` passthrough table. Its leaf keys are flattened into Codex `-c key=value` overrides and applied to every Codex job (review, fix, refine, analyze). Codex applies `-c` overrides as its highest-precedence config layer regardless of `--ignore-user-config`, so only the explicitly listed options are injected; the user's full config is still not loaded.

```toml
[agent.codex.config]
model_provider = "my-custom"

[agent.codex.config.model_providers.my-custom]
name = "My Provider"
base_url = "https://api.example.com/v1"
env_key = "MY_API_KEY"
wire_api = "responses"
```

This flattens to `-c model_provider="my-custom"`, `-c model_providers.my-custom.base_url="https://api.example.com/v1"`, and so on.

## Design notes / risk

- Wired through `applyAgentConfigOverrides` (the shared agent-resolution choke point that already injects Pi's extension), so one wiring point covers all Codex job types.
- roborev's own `-c` flags (skill suppression, sandbox mode, reasoning effort) are emitted **after** the user overrides, so roborev's safety controls win on any key collision.
- Values are TOML-encoded to preserve type (quoted strings, bare numbers/bools, arrays).
- The `config` field is a generic `map[string]any`: set it by editing TOML directly (`roborev config set` does not support nested map keys).

Docs updated in `docs/configuration.md`. Changelog left untouched (release-driven).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -808,6 +808,41 @@ ignore_review_user_config = true   # Pass --ignore-user-config to Codex review j
 
 Set either to `false` to restore the prior behavior if you rely on Codex skill instructions or user-level Codex config during reviews.
 
+#### Custom Codex Config (Model Providers)
+
+`ignore_review_user_config` stops Codex review jobs from loading your
+`~/.codex/config.toml`, which also means a custom `model_provider` and its
+`[model_providers.*]` block defined there are not applied. To inject specific
+Codex options without loading your full user config, define an
+`[agent.codex.config]` table. Every key under it is passed to Codex as a
+`-c key=value` override on **every** Codex job (review, fix, refine, analyze),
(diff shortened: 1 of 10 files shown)


mariusvniekerk (approved):

wesm:
thank you!

PR #902, opened 2026-06-26
Support per-analysis agent config

mariusvniekerk:
## Summary

Add per-analysis-type configuration so built-in analysis runs can pin agent, model, and reasoning independently from the broader review/security workflow defaults.

This supports TOML like:

```toml
[analyze.refactor]
agent = "claude-code"
model = "sonnet"
reasoning = "fast"
```

The resolver keeps existing layering semantics: explicit CLI flags win first, then repo-local analyze config, repo workflow/default config, global analyze config, global workflow/default config, and finally the existing hardcoded default. Analyze jobs still flow through the normal daemon agent resolution and invocation path, so agent-specific settings such as `[agent.codex]` command/config overrides continue to apply after an analysis type picks Codex.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,15 @@ Analysis jobs appear in the review queue. Use `roborev fix` to apply
 open findings later, target a specific job with `roborev fix <id>`, or
 pass `--fix` to apply immediately.
 
+Analysis types can pin their own agent settings in config:
+
+```toml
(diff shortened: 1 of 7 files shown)


mariusvniekerk:
This issue IS valid, though the correct bvehavior is to error, not try to design some piece of insanity around it

PR #903, opened 2026-06-26
Add lookahead review type for time-series look-ahead bias

someben:
roborev ships specialized review types (security, design) that swap the system
prompt to focus the reviewer. Code that works with time-indexed data has a whole
class of defects these miss: look-ahead bias, where a computation uses information
that would not be available at the timestamp it represents. This adds a `lookahead`
review type targeting that failure mode.

`roborev review --type lookahead` selects a new system prompt that holds findings
to a point-in-time burden of proof — it must name the future-dated input, the
earlier value it corrupts, and why that input isn't available yet — across window/
lag direction, train/test ordering, fit-on-full-series statistics, resample/fill
direction, as-of join alignment, timestamp semantics, and label/target leakage.
The type is wired through the same paths as security and design: validation, the
template selector, per-workflow agent/model/backup config (resolved by the existing
reflection helpers, no resolver changes), panel/CI routing, CLI flag and completion,
Claude and Codex skills, and docs. The prompt also names common synonyms (peekahead,
future leakage, temporal leakage) so it triggers regardless of vocabulary.

It is opt-in only: no existing review path changes behavior, and it is not added to
the CI poller's default review-type set. Validated locally with `go test ./...`,
`go vet ./...`, and a regenerated golden prompt.


internal/config/config.go
```diff
@@ -180,6 +180,19 @@ type Config struct {
 	DesignModelThorough   string `toml:"design_model_thorough"`
 	DesignModelMaximum    string `toml:"design_model_maximum"`
 
+	LookaheadAgent         string `toml:"lookahead_agent"`
```
mariusvniekerk: There is another PR taking a more general purpose solution to these config params for particular agents #902

mariusvniekerk (commented):
I like this in general, probably want to wait for #902 to land to have the more general purpose config per analysis agent

mariusvniekerk:
so which agents work well with this one?  I assume using a model with stronger spatial reasoning helps.

someben:
> so which agents work well with this one? I assume using a model with stronger spatial reasoning helps.

Claude's okay, but none of them are that great. A guy can dream.

wesm:
couldn't push here for some reason so I opened #904

PR #904, opened 2026-06-27
Add lookahead review type for time-series bias

wesm:
This adds a dedicated `lookahead` review type for time-series look-ahead bias: cases where code uses information that would not have been available at the point in time being represented. The type is exposed through `roborev review --type lookahead`, CI/panel review type configuration, prompt templates, docs, and bundled agent skills so it can be used the same way as existing specialized review types.

Lookahead is intentionally configured through the generic `[analyze.lookahead]` bridge rather than new `lookahead_*` config fields. That keeps the new type lightweight while still allowing agent/model pinning for review workflows that do not have native config fields.

The config resolver keeps existing `security` and `design` review behavior isolated from legacy `[analyze.security]` and `[analyze.design]` tables, so analyze-only settings do not silently change specialized review jobs.


wesm:
LGTM

PR #915, opened 2026-06-29
Repair managed hooks after update

### LINKED ISSUE #905: Roborev not automatically reviewing commits
Hello, again.

I upgraded from 0.56 to 0.60 
In this new version, roborev isn't picking up commits automatically to review - `roborev tui` shows no new reviews.
I manually call `roborev review HEAD` or I find the commit ID and then --> `roborev review 123456`.

So, at least roborev is working because after manually requesting a review for a commit `roborev tui` shows a new review/JobID, but it doesn't seem to be picking them up automatically.

I should note that before the update in the same repo, roborev worked.

OS: Windows 11 Enterprise
terminal: git bash

cpcloud:
This fixes stale managed Git hooks that can keep invoking an old roborev binary after `roborev update`, causing automatic reviews to be enqueued through the wrong daemon while manual reviews use the upgraded binary.

The update command now runs the hook repair path for all registered repos after an update, passing the freshly installed roborev binary path into the hook writer. The repair logic only rewrites roborev-generated hook blocks identified by their marker comments, so unmanaged hooks and existing hooksPath configuration are left alone.

The manual repair surface is `roborev install-hook repair` for the current repo and `roborev install-hook repair --registered` for all registered repos.

For users on existing releases that do not include this fix, the manual workaround is still to run `roborev install-hook` once from each repo where automatic reviews should run. That rewrites roborev's managed `post-commit` and `post-rewrite` hooks to the currently resolved roborev binary. If you installed roborev through a version manager or shim, use `roborev install-hook --binary <stable-roborev-path>` from each repo so future hook invocations do not stay pinned to a versioned binary.

Closes #905.

### DIFF
--- a/cmd/roborev/hook.go
+++ b/cmd/roborev/hook.go
@@ -53,6 +53,7 @@ func installHookCmd() *cobra.Command {
 
 	cmd.Flags().BoolVar(&force, "force", false, "overwrite existing hook")
 	cmd.Flags().StringVar(&hookBinary, "binary", "", "roborev binary path to bake into git hooks (for version-manager shims)")
+	cmd.AddCommand(installHookRepairCmd())
 
 	return cmd
 }
--- a/cmd/roborev/hook_reconcile.go
+++ b/cmd/roborev/hook_reconcile.go
@@ -1,0 +1,204 @@
+package main
+
+import (
+	"context"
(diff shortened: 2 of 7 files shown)


wesm:
refining

PR #954, opened 2026-07-11
docs: ACP bridge example for model-selectable Gemini, plus config gotchas

mjacobs:
## Summary

- Add an ACP-docs example showing how a bridge (e.g. [agy-acp](https://github.com/mjacobs/agy-acp), which wraps the Google Antigravity SDK) restores explicit Gemini model selection — including a thinking-level suffix — through the existing generic ACP agent, since the `agy` CLI has no model flag and the legacy `gemini` CLI is unavailable to most users
- Document the `env`-wrapper pattern for passing environment variables to daemon-spawned ACP agent processes
- Add two troubleshooting entries observed in real usage: workflow models (`review_model` etc.) overriding the `[acp]` `model` field and causing silent retry+failover to the backup agent, and `default_panel` absorbing `--agent` runs unless `--panel none` is passed

https://claude.ai/code/session_01HwTz3BHkFuESb3BauS7F9r

### DIFF
--- a/docs/advanced/acp.md
+++ b/docs/advanced/acp.md
@@ -67,6 +67,41 @@ args = ["--verbose"]
 
 Once configured, the agent can be selected with `--agent my-agent`.
 
+### Example: Model-Selectable Gemini via a Bridge
+
+The `agy` (Antigravity) CLI does not support explicit Gemini model selection,
+so the built-in `gemini` agent runs default-model-only when it resolves to
+`agy`. An ACP bridge that wraps the Google Antigravity SDK restores model
+selection (including thinking-level control) through the generic ACP agent —
+no roborev changes required. One such bridge is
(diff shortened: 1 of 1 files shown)


mariusvniekerk (approved):

mariusvniekerk:
kinda hilarious that the best way to use antigravity is not via their cli.

mjacobs:
> kinda hilarious that the best way to use antigravity is not via their cli.

:zany_face: ...but it's only the best way if you're using _API keys_ or ~Vertex AI~ _er_ "Gemini Enterprise Agent Platform" (sigh: https://docs.cloud.google.com/gemini-enterprise-agent-platform/vertex-ai-name-changes) because the SDK doesn't support the consumer subscription path, so for _that_ you still wanna use `agy` cli, there's just no way to specify a model :exploding_head: 

snippy_1783849307483


and meanwhile, gemini cli worked for all of these cases... sigh

PR #962, opened 2026-07-12
ACP Gemini follow-ups: backup-path model pairing guard, plus docs

mjacobs:
## Summary

Completes the Gemini-via-ACP story from #954/#955 (these follow-ups were pushed to those branches moments after the squash-merges, so they're re-landed here on top of #961).

**Code:** #955 established that a configured model only applies to the agent it's paired with. This PR extends that same principle to the one path it didn't cover: backup/failover. Previously, a global `default_backup_model` could be inherited by a workflow-selected ACP backup agent it was never paired with — and because ACP agents validate models by exact membership, the mismatch doesn't degrade, it breaks the failover itself. The guard now applies same-layer pairing on both the resolution and daemon-failover paths, keeps `[acp].model` in place on a mismatch (and reflects it in `job.model`), and leaves non-ACP agents and all legitimately-paired backup configs untouched. Pinned by a table-driven test across the pairing layers plus a failover-path test.

**Docs:** rounds out the Gemini guidance in `docs/advanced/acp.md` — a "Which Gemini path should I use?" decision tree keyed on how you authenticate (subscription OAuth vs `GEMINI_API_KEY` vs Vertex), a credentials-safe pattern for passing env to daemon-spawned ACP agents, and precision notes on model validation and `--model`/panel interaction — and aligns `docs/agents/index.md` with the current Antigravity behavior (explicit-model handling, the #952 version-gated prompt contract).

snippy_1783849307483

### DIFF
--- a/docs/advanced/acp.md
+++ b/docs/advanced/acp.md
@@ -88,11 +88,19 @@ Then select it anywhere agents are routable: `--agent agy-sdk`, per-workflow
 agents, backup agents, or panel members. Bridge model IDs accept an optional
 thinking suffix (`gemini-3.5-flash:high`), which is how a reasoning level
 reaches the model — roborev's ACP client transmits only mode and model, not
-reasoning.
+reasoning. The suffixed IDs work only because the bridge advertises the
+suffixed strings verbatim in its model list: roborev validates the configured
+model against exact membership of what the agent advertises (for agy-acp, the
+set in `AGY_ACP_MODELS`), so a suffix the bridge does not advertise is
+rejected just like any unknown model.
 
 If the agent process needs environment variables (API keys, cloud project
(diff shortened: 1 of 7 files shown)


mjacobs:
Addressed the CI review's Medium in f67b47d: backup models now pair with the backup agent resolvable at the model's own layer — repo-layer models pair with the normally-resolved agent, a global `{workflow}_backup_model` pairs with the global-layer agent resolution only, and a global `default_backup_model` pairs with `default_backup_agent` only — so a repo-level `review_backup_agent` can no longer capture a global workflow backup model. Coverage added for the repo-override + global-model mispair, the legit cross-file and same-file pairs, the non-ACP scope guard, and the daemon failover path (which already routes through the same guarded chokepoint).

PR #965, opened 2026-07-12
Remove flake-utils from Nix flake

luisnquin:
This removes flake-utils and replaces eachDefaultSystem with a local system list plus nixpkgs.lib.genAttrs. This project only needs a small fixed set of outputs, so the extra flake-utils dependency is not needed.

Removing it also drops the transitive systems input from flake.lock, reducing flake weight for external consumers.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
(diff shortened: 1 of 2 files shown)


wesm:
thank you

PR #972, opened 2026-07-14
feat: add machine-readable run launch receipts

iamcxa:
## What changed

- Add `roborev run --json` for a machine-readable launch receipt after a review job is accepted.
- Emit exactly one JSON object with `{job_id,job_uuid,git_ref,status}`.
- Add a dedicated enqueue response schema and generated client type for the launch boundary.
- Cover the CLI behavior, JSON shape, and HTTP client decoding with tests.

## Why

Automation needs an atomic, stable receipt that identifies the job it just launched. The existing command prints a human-oriented enqueue message, which requires callers to parse presentation text and cannot serve as a durable machine contract.

The root cause is that the enqueue endpoint and CLI did not expose a narrow launch-receipt contract. Reusing the general `ReviewJob` response would also couple enqueue callers to a broader model whose compatibility requirements can evolve independently.

## Impact and design choices

With `--json`, callers receive:

```json
{"job_id":1,"job_uuid":"example-uuid","git_ref":"example-ref","status":"queued"}
```

Normal human-readable output remains unchanged when `--json` is absent. The implementation uses a dedicated enqueue schema so the existing general `ReviewJob` client contract remains compatible.

The following alternatives were intentionally rejected:

- `--quiet`: suppresses output instead of providing structured identity.
- `--wait`: changes lifecycle behavior and does not solve the launch-receipt boundary.
- global `--verbose`: is intended for diagnostics, not stable machine-readable data.

No real provider job was launched while validating this change.

## Validation

- `go test ./... -count=1`
- `make lint-ci` using the pinned `golangci-lint` version
- `go vet ./...`
- `go build ./...`
- `make api-generate`
- `git diff --check`

### DIFF
--- a/cmd/roborev/run.go
+++ b/cmd/roborev/run.go
@@ -24,14 +24,15 @@ import (
 
 func runCmd() *cobra.Command {
 	var (
-		agentName string
-		model     string
-		reasoning string
-		wait      bool
-		quiet     bool
-		noContext bool
-		agentic   bool
-		label     string
+		agentName  string
+		model      string
+		reasoning  string
+		wait       bool
+		quiet      bool
+		noContext  bool
+		agentic    bool
+		label      string
+		jsonOutput bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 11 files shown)


wesm:
looking at this

wesm:
superseded by #975 which I just merged

PR #975, opened 2026-07-14
Add machine-readable run launch receipts

wesm:
## Summary

- Add a `--json` flag to `roborev run` that emits one machine-readable launch receipt (`job_id`, `job_uuid`, `git_ref`, `status`) from the atomic enqueue response; incompatible with `--quiet`, `--wait`, and the global `--verbose`
- When the daemon skips the enqueue (e.g. excluded branch), `--json` emits a single `{"skipped":true,"reason":...}` document and human mode prints the reason
- Return a dedicated `EnqueueCreatedResponse` (required `uuid`) from the created enqueue path and document it in the OpenAPI spec alongside `PanelEnqueueResponse` and `EnqueueSkippedResponse` in the success oneOf; the general `ReviewJob` schema keeps `uuid` optional for older stored records
- Stop huma's `ReviewJob` link transformer from injecting an undeclared `$schema` field into 201 enqueue bodies, and pin that with a test
- Regenerate `pkg/client` and `internal/daemon_client` from the updated spec

### DIFF
--- a/cmd/roborev/run.go
+++ b/cmd/roborev/run.go
@@ -24,14 +24,15 @@ import (
 
 func runCmd() *cobra.Command {
 	var (
-		agentName string
-		model     string
-		reasoning string
-		wait      bool
-		quiet     bool
-		noContext bool
-		agentic   bool
-		label     string
+		agentName  string
+		model      string
+		reasoning  string
+		wait       bool
+		quiet      bool
+		noContext  bool
+		agentic    bool
+		label      string
+		jsonOutput bool
 	)
 
 	cmd := &cobra.Command{
@@ -49,6 +50,10 @@ The task can be provided as:
 By default, the job is enqueued and the command returns immediately.
 Use --wait to wait for completion and display the result.
 
+Use --json to emit one machine-readable launch receipt containing job_id,
+job_uuid, git_ref, and status. It cannot be combined with --quiet, --wait,
+or the global --verbose flag.
+
 By default, context about the repository (name, path, and any project
 guidelines from .roborev.toml) is included. Use --no-context to disable.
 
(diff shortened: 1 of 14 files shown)


wesm:
A code review of the original two commits surfaced nine findings (six confirmed by adversarial verification, plus measured convention breaches). Commit `ca6b0504` addresses all of them:

**Correctness / contract**

1. **`--json` broke on skipped enqueues.** Branch exclusion is checked before job-type dispatch, so a `roborev run --json` on an excluded branch got a 200 `{"skipped":true,...}` body that the CLI turned into a prose error (`enqueue failed: {...}`) with exit 1 — violating the single-JSON-document contract. Now both modes handle it: `--json` emits the skipped document, human mode prints the reason (same pattern as `roborev insights`), exit 0. Covered by two new subtests.
2. **`internal/daemon_client` was stale.** The enqueue oneOf change was regenerated only into `pkg/client`; `go generate ./internal/daemon_client` produced an 821-line diff. Both generated clients are now in sync with the daemon spec.
3. **The documented schema rejected real responses.** Actual 201 bodies carried a huma-injected `"$schema"` field (from the `ReviewJob` link transformer via the fix-job route), while `EnqueueCreatedResponse` declares `additionalProperties: false` without it — every successful enqueue would fail a strict spec validator. Fixed by the change below; a test now pins that 201 bodies carry no `$schema`.
4. **`EnqueueCreatedResponse` was documentation-only.** The handler still returned `*storage.ReviewJob`, so nothing but a test tied the payload to the required-uuid schema, and the shadow `UUID` field would silently serialize `""` if ever instantiated without it. The created path now returns `EnqueueCreatedResponse{ReviewJob: job, UUID: job.UUID}` — the compiler enforces the contract (mirroring `PanelEnqueueResponse`).
5. **Panel enqueues weren't in the documented union.** `POST /api/enqueue` can return 201 `PanelEnqueueResponse`, whose `member_job_ids` matched neither oneOf member (gap predates this branch). The success oneOf now has three members; the generator emits a raw-message union with `As<Type>()` accessors.

**Robustness / cleanup**

6. **Receipt validation fired after the job was enqueued and hid that fact.** The git_ref/status checks were unreachable against any same-version daemon (storage assigns all three at insert). Collapsed to a single uuid guard whose error states the task *was* enqueued and points at daemon version skew, so automation doesn't retry and duplicate jobs.
7. **Flag-conflict errors** now go through the repo's `usageErr` convention instead of a hand-rolled `SilenceUsage` toggle (redundant with root's `PersistentPreRunE`).
8. **`runPrompt` had grown to 118 lines / 12 positional params.** Refactored to a `runOptions` struct with `validateRunFlags`/`readRunPrompt`/`reportRunSkipped` helpers; now 95 lines, 3 params.
9. **Test/doc cleanup:** the re-query subtest reuses `newRunTestServer` instead of duplicating ~40 lines of inline handlers, and the user-facing docs no longer explain OpenAPI schema internals.

Verified with `go vet`, `golangci-lint` (0 issues), and full test runs of `cmd/roborev`, `internal/daemon`, and `pkg/client`.

Issue #577, opened 2026-03-24
GitLab Integration


amilosevic-grid:
From what I see, for github integration, gh CLI tool is used. 
In a similar way GitLab can be integrated  in a similar way using the glab CLI tool as well.

I have been using roborev for a while and I use it for personal projects that I put on Github, but I would also love to integrate it on the CI for company projects on GitLab.

Would you accept a PR on this? @wesm

wesm:
Sure, I think the exiting CI poller infrastructure should be able to be used to provide the same kind of experience

nicoa:
@amilosevic-grid I hope it's fine for you if I take over to make a first attempt on a PR 🙏 will try my luck and file one soon.

nicoa:
@wesm @amilosevic-grid wall of text incoming from AI 😀  **TL;DR: I implemented only the roborev ci review (for now) in the linked MR** - please find details below

---

Opened #1004 for the first slice of this: GitLab support in roborev ci review (pipeline mode), which is the daemon-less path that runs inside a CI job and posts the result as an MR note. It's running in production on an internal GitLab project, reviewing its own merge requests.

Scope choice. I split this into three PRs rather than one: 
1) pipeline mode, which is self-contained and needs no daemon/storage changes
2) extracting a forge interface from the CI poller's existing test seams as a behaviour-neutral refactor
3) GitLab in the daemon poller. 

#1004 is only the first. That keeps the reviewable surface small and gets the CI use case working before touching poller internals.

Tradeoffs I picked, and why:

- Official gitlab.com/gitlab-org/api/client-go rather than a hand-rolled REST client. The repo prefers stdlib, and ~8 endpoints wouldn't have justified a dependency — but it mirrors the existing go-github choice, and swapping it later is contained. Happy to revisit if you'd rather keep the dependency count down.
- Marker-comment convention shared with GitHub, pinned by a test asserting the marker strings are equal, so upsert behaviour and comment formatting can't drift between forges.
- Quick-action escaping is unconditional, including inside code fences. GitLab executes lines starting with /word in note bodies as quick actions, so a prompt injection in an MR could have made a synthesized review run /approve on itself. I first implemented fence-aware escaping, but every attempt to mirror GitLab's fence semantics left an arrangement where our parse and GitLab's disagreed, and each disagreement restores a live /approve. The cost is a cosmetic \/ inside code samples; I judged that worth removing the whole bypass class. Verified against GitLab's extractor source rather than the docs.
[...]

PR #1004, opened 2026-07-28
Add GitLab merge request support to roborev ci review

nicoa:
## Summary

- Add `internal/gitlab` package wrapping the official `gitlab.com/gitlab-org/api/client-go`: token resolution (`GITLAB_TOKEN`/`GL_TOKEN` env vars with `glab auth token` fallback), self-hosted base-URL support, and MR note create/upsert using the same marker convention as the GitHub PR comments
- Support GitLab merge requests in `roborev ci review`: new `--gl-repo` flag (mutually exclusive with `--gh-repo`) and auto-detection inside GitLab CI (`CI_PROJECT_PATH`, `CI_MERGE_REQUEST_IID`, ref range from `CI_MERGE_REQUEST_DIFF_BASE_SHA..CI_COMMIT_SHA` with fallbacks); the GitHub path is unchanged
- Generalize the test API guard to a blocked-host list covering gitlab.com alongside api.github.com
- Add `docs/integrations/gitlab.md` with a sample `.gitlab-ci.yml` and token requirements (`CI_JOB_TOKEN` cannot create MR notes — a project access token is required); extend the CI Review section in `docs/commands.md`

First slice of #577 (pipeline mode only). Follow-ups planned: extracting a forge interface from the CI poller's existing test seams, then GitLab support in the daemon poller (config keys, `gitlab:`-prefixed storage keys, MR semantics mapping).


nicoa:
Hint / comment: I'll test this throughout the next days on a private gitlab repo, making this a draft

nicoa:
Addressed all three findings:

- **High — token exposure in MR pipelines**: reworked the token guidance around an explicit trust model (6a24635 docs part in 0f7e03b): protected+masked project access token as the recommended default, with the protected-branch posting flow documented (`--pr <iid> --ref <base>..<head>` plus the required `git fetch origin refs/merge-requests/<iid>/head`), and the unprotected variant explicitly limited to internal projects with trusted authors. Personal/group tokens are no longer suggested as defaults.
- **Medium — synthetic merge commit in merged-results pipelines**: the head of the auto-detected range now prefers `CI_MERGE_REQUEST_SOURCE_BRANCH_SHA`, falling back to `CI_COMMIT_SHA` (6a24635, with tests).
- **Medium — fork pipelines posting to the wrong project**: project auto-detection now prefers `CI_MERGE_REQUEST_PROJECT_PATH` (the MR's target project), falling back to `CI_PROJECT_PATH` (6a24635, with tests).

Docs follow-ups in 0f7e03b/5bd0d20/12d7246 keep the setup instructions consistent with the example pipeline.

nicoa:
Addressed both findings:

- **Example pipeline not runnable**: the example now uses `node:22-bookworm` (ships git + Node) and installs `@anthropic-ai/claude-code` in `before_script`; prose spells out the actual requirements (git for the `base..head` diff, the agent CLI's runtime, the CLI on `PATH`) so other images/agents are covered too (50862a1).
- **`ANTHROPIC_API_KEY` not reaching Claude Code**: real code gap, not just a docs issue — `ci review` never called `agent.SetAnthropicAPIKey`, so the adapter's env sanitization left claude-code without credentials in pipeline mode. `runCIReview` now resolves the key (global-config `anthropic_api_key` first, `ANTHROPIC_API_KEY` env fallback, mirroring `refine.go`) and hands it to the agent before the batch runs, with table-driven tests (88c3fae). The docs now describe the actual mechanism. This also makes the GitHub Actions pipeline docs' existing claim accurate — same code path.

Codex/Gemini agents inherit the job env unchanged (only the Claude adapter filters), noted in the docs.

nicoa:
Both fixed.

**Quick-action injection** (`111ed8b`, `21c2f99`, `ab488d5`, `e9c9da7`): verified the mechanism against GitLab's own extractor (`lib/gitlab/quick_actions/extractor.rb`) rather than the docs — the command branch is anchored `^/` (column 0 only) and the body has every `\r` deleted before matching. `prepareBody` — which both the create and upsert paths already funnel through — now escapes any line matching `^/[a-zA-Z]` after CR stripping, inserting `\` before the slash so GitLab's parser no longer matches and CommonMark still renders a literal `/`. Idempotent, so re-upserting an existing note doesn't double-escape.

Notably, escaping is **unconditional** rather than skipping fenced code blocks. I implemented fence-awareness first, but every attempt to mirror GitLab's fence semantics left another arrangement where our parse and GitLab's disagreed — asymmetric open/close rules, an empty fence pair not counting as a block (shifting all later boundaries), and the HTML/`>>>` quote branches swallowing a fence opener — and each disagreement restores a live `/approve`. Unconditional escaping costs a cosmetic `\/` inside code samples and removes the whole bypass class; it also let the fence-closing-on-truncation workaround go away. Happy to revisit if you'd rather keep fenced samples pristine, but that re-accepts the bypass.

Tests: 20-case table (`/approve`, `/close`, `/target_branch foo`, mid-line slashes, URLs, indented, `/usr/bin/foo`, fenced, adversarial empty-fence and HTML-wrapped bodies, leading/interior CR, already-escaped), each also asserting idempotency, plus create/upsert path and truncation coverage.

**Fallback diff range** (`4c6a572`, `c34bcbd`): when the base comes from `CI_MERGE_REQUEST_TARGET_BRANCH_SHA` or `CI_COMMIT_BEFORE_SHA`, the range base is now `git merge-base <base> <head>` via the existing `git.GetMergeBase`. `CI_MERGE_REQUEST_DIFF_BASE_SHA` is still used verbatim since GitLab already resolved it. Merge-base failure (shallow clone, pruned objects) logs a warning and keeps tip-to-tip rather than failing the job. `CI_COMMIT_BEFORE_SHA` gets the same treatment deliberately — it stops being an ancestor of head after a force-push or rebase, so it has the same bogus-removals problem. Ref detection moved after repo-root resolution to make the merge-base call possible; flag validation still runs first. Diverged-branches tests cover both fallback vars.

nicoa:
**High — forge token reachable by review agents: fixed** (`d6d29a7`).

Agents no longer inherit forge credentials. `GITLAB_TOKEN`, `GL_TOKEN`, `CI_JOB_TOKEN`, `GH_TOKEN` and `GITHUB_TOKEN` are stripped from every agent subprocess environment at a single choke point (`configureSubprocess` in `internal/agent/process.go`), so a newly added agent can't forget to do it. roborev resolves these credentials in its own process, so comment posting and every other API call are unaffected.

Two deliberate exceptions, both documented in `internal/agent/forge_env.go`:

- `copilot` and `kiro-cli` authenticate *with* a GitHub token and cannot run without one, so those two launch sites opt back in via `withGitHubCredentials()`. Every other agent gets them removed.
- `GOOGLE_APPLICATION_CREDENTIALS` is **not** stripped — Claude Code on Vertex AI authenticates with it, so removing it would break that path.

`CI_JOB_TOKEN` is included even though roborev never uses it (it can't create notes): GitLab injects it into every job, it grants API access for the job's lifetime, and no agent has a legitimate use for it.

Worth noting the exposure this closes is real but narrower than "shell access" — review mode runs Claude Code with `--allowedTools Read,Glob,Grep` and no `--dangerously-skip-permissions`, so there's no Bash tool. The live channel was the one your fix note implies: an injected agent copying a token it can read into the review text, which then gets posted as a comment. Stripping the environment closes that regardless of tool surface.

**Medium — stale tree: partly accurate, now documented** (`4b356a1`, `94d8344`).

Correct for the protected-branch flow the docs describe: `git fetch origin refs/merge-requests/<iid>/head` makes the commits reachable but doesn't move the working tree, so the `<base>..<head>` range is right while agents opening files for context read the protected branch's copies.

Not correct for a normal merge request pipeline, which is the common setup: the runner has already checked out the MR branch — or, in a merged results pipeline, the simulated merge — so the files agents read already contain the change. `RunBatch` inherits whatever the job checked out; it isn't pinned to the default branch.

Rather than have `ci review` create worktrees itself (a larger design change than this PR should carry), the protected-branch section now states the gap explicitly and shows the remedy: `git worktree add ../mr FETCH_HEAD` and review from there. A follow-up commit notes the consequence — reviewing from the MR-head worktree means `.roborev.toml` is read from that tree, so the merge request's own config governs its review.

Separately, filed #1005 for a related exposure this PR surfaced but doesn't introduce: in pipeline mode the agent's working directory is the untrusted PR/MR checkout, and Claude Code loads project config from there — hooks in a committed `.claude/settings.json` execute with no trust prompt under `-p`, which a read-only tool allowlist doesn't cover. It affects the GitHub Actions path equally.

nicoa:
**Windows case-sensitivity: fixed** (`78c3354`).

Correct and a real bug. Windows resolves environment variable names case-insensitively, so a credential set as `GitLab_Token` is readable via `os.Getenv("GITLAB_TOKEN")` — roborev would still resolve it for API calls while the exact-match filter left it in the agent's environment. `filterEnv` now folds case when the platform does, behind a test seam (`envKeysAreCaseInsensitive`) so both behaviours are exercised on any host. Tests cover mixed-case GitLab and GitHub tokens, the canonical spelling, the POSIX case where a different case genuinely *is* a different variable, and provider credentials surviving in both modes.

**`glab auth token` fallback: not changing it here, and I'd argue not in this PR.**

The mechanism is real — an agent with shell access can invoke `glab auth token` and reach the host credential store no matter what the environment filter removes. Two reasons it's the wrong edit for this change:

1. `internal/github/client.go:31-47` has the identical `gh auth token` fallback, with the same purpose and the same exposure. I mirrored it deliberately, and the GitLab docs describe it as the local-run convenience it is. Removing it only on the GitLab side would make the two forges behave differently for the same operation, which is a worse outcome than the shared risk.
2. It isn't reachable on the path this PR adds. In CI `GITLAB_TOKEN` is set, so the fallback never runs — it exists for invoking `roborev ci review` from a workstation. And review mode launches Claude Code with `--allowedTools Read,Glob,Grep` and no `--dangerously-skip-permissions`, so there is no shell to run `glab` with. The finding really concerns agents whose tool surface roborev does not constrain as tightly, which is orthogonal to the forge.

So this reads as a general question about what a review agent can reach in an untrusted-review context, not something the GitLab client introduces. I've added it to #1005, which already covers that theme (project config and hooks loaded from the untrusted checkout) and applies to both forges. If you'd rather the credential-store fallback be gated on something explicit — say, only consulted outside CI, or behind a config flag — I'm happy to do it for `gh` and `glab` together in a follow-up so the two stay symmetric.

nicoa:
On the High finding about `CI_SERVER_URL` and the API origin (`internal/gitlab/client.go:140`): the mechanism is real, but we're accepting it as a documented limitation rather than fixing it in code. Documented in 2d08a44.

**What checks out.** `DefaultServerURL` does trust `CI_SERVER_URL` first, then `GITLAB_HOST`/`GL_HOST`. `ciGitLabClient` always calls `GitLabAPIBaseURL("")`, so the origin only ever comes from the environment — there is no flag or config key to pin it. The token is then attached to whatever that resolves to. And predefined variables really are last in GitLab's variable precedence, so a pipeline started with custom variables can override `CI_SERVER_URL`. All four verified.

**Why not High here.** Doing this requires starting a pipeline on a protected branch with custom variables, which needs push or merge rights on that branch, or a pipeline trigger token. Those actors can already obtain the token by simpler routes. The Trust Model section of `docs/integrations/gitlab.md` already states the governing property — anyone who can run a pipeline can read the variables exposed to it — so for the unprotected-variable setup this is an instance of an already-documented property rather than a new exposure. The code does what it documents; this is a hardening gap, not a defect.

**The case we agree has teeth**, and which the docs now call out explicitly: a Developer allowed to merge to a protected branch whose merge requests still require approval. They can start a protected pipeline with variables without landing any code, which turns "can trigger" into "can exfiltrate." Projects in that configuration should treat the ability to trigger protected pipelines as equivalent to holding the token.

**On the suggested fix.** "Pin the origin in trusted CI configuration" isn't available today — it would need a new `--gl-host` flag plus a config key, so accepting this means accepting that an operator who wants the control has none. We're taking that trade for now and will add the pin if someone runs the protected-variable flow with semi-trusted Developers. Worth recording for whoever implements it: a repo-config anchor is only trustworthy in the protected-branch flow, because in the MR-worktree flow documented above `.roborev.toml` comes from the author's tree — the flag would have to be the anchor.

The warning against using another overridable predefined variable as the trust anchor was well taken: `CI_SERVER_HOST` and `CI_SERVER_FQDN` would have been security theater, and that's noted in the docs so nobody reaches for them later.

nicoa:
Two High findings on `2d08a44`: fixed the first in 335a029, rejecting the second as already adjudicated.

## Incomplete reviews in shallow merge-request checkouts — accepted, and it was ours

Correct, and it was a defect in the shallow-clone guard added in 8645a37 rather than a pre-existing one. Confirmed the mechanism: a single ref is diffed with `git show` (`internal/git/git.go:789`), so returning `head` alone reviewed a twelve-commit merge request as its last commit — and the run could still post "Review Passed" on that basis. Trading a loud failure for a silent scope reduction was the wrong call.

The guard had collapsed two different situations, and 335a029 splits them:

- **No base variable set at all** — a plain branch pipeline. There is no merge request range, so the HEAD commit is the whole review rather than a narrowed one. Unchanged; this behavior predates the guard.
- **A base was named but is not in the clone** — now errors, naming the SHA, the variable it came from, and `GIT_DEPTH`. `base..head` could not have worked either, since every later git operation would hit the same missing commit, so failing is the only honest option.

This also makes `detectGitLabGitRef` self-consistent: the head guard added in 75df3f5 already errored rather than degrading when no candidate head was present, while the base guard a few lines below did the opposite.

`TestDetectGitLabGitRef_MissingBaseReviewsHeadAlone` asserted the behavior being removed, so it is now `_MissingBaseErrors` and checks that the message names the SHA and `GIT_DEPTH`; `_NoBaseReviewsHeadAlone` pins the branch-pipeline path. Verified load-bearing — restoring the head-only return fails all three subtests with "An error is expected but got nil".

## GitLab token exfiltration through an overridden API origin — already adjudicated

This is the same finding as the previous round, raised at the caller (`ci.go:841`) instead of the resolver (`client.go:140`). Same mechanism, same suggested fix, same preconditions, no new information — so the decision stands: accepted as a documented limitation, with the reasoning in https://github.com/kenn-io/roborev/pull/1004#issuecomment-5129477646 and the exposure written into the Trust Model section of `docs/integrations/gitlab.md`.

Worth noting for anyone reading the thread: this was flagged while reviewing `2d08a44`, the commit whose entire content is documenting the acceptance. That suggests these reviews are diff-scoped and do not carry the prior acceptance as context, so it will likely resurface on later commits. That is not a reason to revisit the decision, and we would rather not re-litigate it each round — if the call should change, the trigger is a project running the protected-variable flow with Developers who can trigger protected pipelines, which is the one case with real teeth.

nicoa:
Findings on `335a029`: fixed the Medium in ff971d5; the High is the same origin issue for the third time.

## First-push pipelines may review only the final commit — accepted, and it caught a bad justification

Correct, and it exposed reasoning I got wrong one commit earlier. 335a029 kept head-only for the "no base variable at all" case on the grounds that it is a plain branch pipeline with no range to cover. That does not hold where the code is actually reached: merge request pipelines always set the diff base, and later pushes set a non-zero `CI_COMMIT_BEFORE_SHA` — so an empty base means a **first push**, which usually carries several commits. Since a single ref is diffed with `git show`, a five-commit first push was reviewed as its last commit and could still post a verdict. That is the same defect class 335a029 set out to fix, sitting in the branch that commit declared correct.

`gitLabFirstPushRef` now derives the range from the merge base with `CI_DEFAULT_BRANCH`, preferring `refs/remotes/origin/<branch>` since a CI checkout often lacks the local branch. Two edges:

- Head already contained in the default branch — the push added nothing on top, so the commit itself is the whole review.
- Default branch unset or absent from the clone — no way to tell a one-commit push from a twelve-commit one, so it fails and asks for an explicit `--ref` rather than review a fraction and imply the rest passed.

Taking your first suggestion with the second as the fallback, since the merge base gives a correct range when the history is there and the error is the honest answer when it is not.

One thing worth stating rather than glossing: `CI_DEFAULT_BRANCH` is overridable like every predefined variable. Here it only selects a review scope, and whoever can override it already chose what the branch contains — so it is not the trust-anchor problem an overridable API origin would be. That distinction is in the function comment so nobody later mistakes it for treating a pipeline variable as trusted.

Three table cases asserted the removed behavior and now expect the error. `PresentSourceSHAStillWins` was leaning on the no-base path, so it gets a diff base to keep its assertion about head preference intact. Verified load-bearing — restoring head-only fails the new first-push tests.

## Overridable GitLab API origin — third raise, no new information

Same mechanism, same suggested fix, same preconditions as the two previous rounds. Adjudicated and documented; the reasoning is at https://github.com/kenn-io/roborev/pull/1004#issuecomment-5129477646 and the exposure is written into the Trust Model section of `docs/integrations/gitlab.md`.

As predicted last round, this is resurfacing on every commit, which is consistent with the reviews being diff-scoped and not carrying prior adjudications. We will not keep replying to it individually. If the decision should change, the trigger is a project running the protected-variable flow with Developers who can trigger protected pipelines — the one case with real teeth — and the fix at that point is a pinned `--gl-host` plus origin allowlist, not a cross-check against another predefined variable.

nicoa:
Hey @mariusvniekerk @wesm what is your stance on this - should we bring it live and accept the findings (imho, they are acceptable lower risks) or iterate further on it?

wesm:
Let me have a look at it

wesm:
Pushed 16292b3e to address the two review findings.

**Protected token redirect (High).** Confirmed real, but only for the protected-variable setup, and worth spelling out why it differs from the GitHub path even though `ciGitHubClient` uses the same env-derived pattern. In GitHub Actions, job env comes from the runner or from committed workflow YAML, so anyone who can redirect `GITHUB_API_URL` already controls job code running with the secret — the override adds no new capability. In GitLab, pipeline variables outrank predefined variables and can be injected at pipeline start (web UI, API, trigger tokens) without touching any code, so "can start a protected-branch pipeline" and "can run job code with the token" are separable. A trigger-token holder — who can otherwise only start pipelines — could set `CI_SERVER_URL` and capture the `api`-scope token. Since the recommended protected flow is driven by triggers/webhooks, that path was load-bearing.

Changes:
- New `--gl-host` flag that outranks `CI_SERVER_URL`/`GITLAB_HOST`/`GL_HOST`. Hardcoded in the job script it is a real pin: the script comes from the ref's `.gitlab-ci.yml`, which pipeline variables cannot rewrite. (A protected CI/CD variable would not work — pipeline variables outrank project variables too.)
- `GitLabAPIBaseURL` now rejects plaintext `http` origins (loopback excepted, so local stubs keep working), URLs embedding credentials, and non-http(s) schemes, on every resolution path.
- Trust Model docs rewritten around the pin: the trigger-token escalation is described, both protected-flow examples pass `--gl-host`, and expanding a variable there (`--gl-host $CI_SERVER_URL`) is called out as defeating the point.

**Initial-push scoping (Medium).** Narrower than stated: `merge-base(default, head) == head` means every pushed commit is already contained in the default branch, so for a feature branch there is nothing new to review and returning head alone is fine. The real gap was the default branch's own first push (a brand-new repository), where the merge base with itself is head and a multi-commit initial push was silently reviewed as its tip. Now a single-commit initial push reviews that commit, and a multi-commit one fails with an actionable message suggesting `--ref "$(git rev-list --max-parents=0 HEAD)..HEAD"`, consistent with the branch's existing fail-rather-than-narrow behavior. Added a `git.CommitCount` helper for this.

Tests: origin-rejection table plus loopback-allowed cases; a flag-beats-env test where a decoy server asserts the env-derived origin is never contacted; both first-push scenarios; `CommitCount` units.

wesm:
Triage of the review on 16292b3: one finding fixed in ba910cad, two rebutted as inconsistent with the threat model this PR establishes.

**Fixed — shallow clone masking a multi-commit initial push (Medium).** Reproduced before fixing: a `--depth 1` clone of a two-commit repo reports `git rev-list --count HEAD` = 1, so the first-push guard would have reviewed the tip. The suggested discriminator (inspect the raw commit object) is the right one — and the obvious alternative doesn't work: a depth-1 clone of a genuinely single-commit repo is also marked `--is-shallow-repository`, so a shallowness check would reject the legitimate case. New `git.IsRootCommit` reads the raw object via `cat-file`; a grafted boundary commit still records its `parent` line, a true root doesn't. `count == 1` is now accepted only for a real root, otherwise the run fails with the existing `GIT_DEPTH` guidance. Tests cover the shallow multi-commit failure and the shallow single-commit case that must keep working.

**Rebutted — token redirect via `CI_SERVER_URL` (High).** This restates the finding 16292b3 addressed without engaging with the fix. The proposed remedy — require a trusted origin whenever an environment token is used — would break the documented unprotected quickstart (`roborev ci review --comment` in an MR pipeline), where env-derived auto-detection is the point and a redirect adds no capability: the MR author already controls the job script and can exfiltrate the token directly. roborev cannot detect whether the token variable is protected, so mandatory pinning would punish the safe majority case to enforce what the docs already prescribe for the protected case: `--gl-host` hardcoded in the job script (which pipeline variables cannot rewrite), plus rejection of plaintext-http and userinfo origins on every resolution path. Validating against runner metadata does not help either — `CI_SERVER_HOST`/`CI_SERVER_FQDN` are predefined variables and equally overridable, which the Trust Model section already notes. The residual (a protected-setup operator ignoring the documented pin) is accepted and documented.

**Rebutted — `CI_JOB_TOKEN` in `.git/config` (High).** Documented as an accepted residual in the Trust Model section, with the operator-side fix named: the runner's `FF_GIT_URLS_WITHOUT_TOKENS` feature flag, or stripping credentials from the remote before the review runs. Having roborev rewrite the checkout's git config would violate the project constraint against mutating user git state — `ci review` also runs on developer laptops, where rewriting `origin` is destructive — and would break later pipeline steps that push through the credentialed remote. It would also be inconsistent with the existing GitHub support: `actions/checkout` persists credentials in `.git/config` the same way, and roborev has never sanitized those checkouts. Refusing to proceed would fail every default-configured runner. `CI_JOB_TOKEN` is additionally job-lifetime and cannot create MR notes, which bounds the blast radius. Environment stripping bounds what a prompt-injected agent reaches through the obvious channel; the checkout itself is governed by the documented operator choice.

wesm:
Triage of the review on ba910ca: the High finding is valid and fixed in 16c1527; the Medium is declined.

**Fixed — forge selection bypassing the origin pin (High).** Confirmed: `detectCIForge` did not consult `--gl-host` and checked `GITHUB_ACTIONS` before `GITLAB_CI`. Since GitLab pipeline variables outrank predefined ones, a pipeline starter could inject `GITHUB_ACTIONS=true` and steer the run to the GitHub client, whose API origin (`GITHUB_API_URL`) and token come from equally injectable variables — sidestepping the `--gl-host` pin entirely. The precondition (a GitHub token in a GitLab job) is realistic here, because the credential strip deliberately keeps `GH_TOKEN`/`GITHUB_TOKEN` for the `copilot` and `kiro` agent CLIs.

Two changes, both test-driven:
- `--gl-host` now selects GitLab at the explicit-flag tier (it exists to pin against injected variables; a pin that injected env can route around is not a pin) and is mutually exclusive with `--gh-repo`.
- The env tiebreak flips to `GITLAB_CI` first. The two indicators are not equally trustworthy when both are set: GitHub Actions job env comes from the runner or committed workflow code, so `GITLAB_CI` inside real Actions can only be set deliberately by trusted code, while `GITHUB_ACTIONS` inside real GitLab CI can be injected by whoever starts the pipeline. This is a deliberate change to the previously documented precedence; the docs and the test that pinned the old order are updated with the rationale.

**Declined — `mergeBase == head` reviews an existing commit (Medium).** When the merge base of the default branch and head equals head, the pushed branch contains zero commits not already on the default branch — nothing is narrowed or missed, which is what the fail-rather-than-narrow guards protect against. Reviewing the commit the branch points at is deliberate, pinned by a test. The suggested "skip and report no changes" would require plumbing a no-op-review pathway through `runCIReview` and the batch runner for a rare corner case (pushing a branch that points at an existing default-branch commit) with no security consequence: this only occurs in branch pipelines, where no MR IID auto-detects, so `--comment` cannot post — the redundant review lands in CI logs only. If the redundancy bothers anyone in practice, it is a fair standalone follow-up, not a blocker here.

wesm:
Triage of the review on 74bbd38: the core of the finding is real and fixed in e9f7641, though the recommended remedy needed narrowing.

**Confirmed:** `isConfiguredACPAgentName` checked the configured `[acp].name` before the agent registry, and `ResolveACPAgentConfig` lets a repo's `.roborev.toml` override global config outright. A repo-local `[acp] name = "claude-code", command = "..."` therefore redefined what `--agent claude-code` executes. In the protected-branch worktree flow this is reached with the MR head checked out, so an author could run an arbitrary command in the protected job — the flag pin covered *which agent name*, not *what that name maps to*. A test reproducing it failed against the previous head and passes now.

**Fix:** a repo-local `[acp]` entry whose `name` collides with a registered agent, directly or through an alias, is ignored — the built-in agent runs instead.

**Where the recommendation needed adjusting.** "Explicit agent flags should resolve strictly to canonical agents without repo-defined ACP aliasing" would also have banned the same shadowing from `~/.roborev/config.toml`, which is a documented, deliberate feature: routing a name like `codex` through an ACP transport (`[acp] name = "codex", command = "acp-agent"`). An existing test pins it, and my first attempt broke that test — which is what surfaced the distinction. The trust boundary is the config's *source*, not the aliasing mechanism: global config is the operator's own file, while `.roborev.toml` comes from the tree under review. The fix restricts only the untrusted source, so both properties hold. `docs/advanced/acp.md` documents the restriction.

The second half of the recommendation — loading batch and synthesis agent config from the default branch instead of the MR worktree — is not adopted here. It is a broader change to config resolution affecting every flow, and the escalation this finding identified is closed by the name restriction. The remaining author-controlled `[ci]` settings (agent choice, review types, `min_severity`) are already documented in the trust model, with the mitigation being to pin them as flags, which the protected-flow example does. Worth a standalone issue if the default-branch config source is wanted generally.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on 08174bf: two findings fixed in bf1d65fb, one declined with the residual documented.

**Fixed — repository config overrides the CLI severity threshold.** Confirmed by reading the call site: `runCIReview` passed `""` as the explicit argument to `ResolveReviewMinSeverity`, so `--min-severity` reached only the synthesis filter while the member-review prompts took `review_min_severity` from the checkout. In the MR-head worktree flow that file belongs to the author under review, who could raise it to `critical` so medium and high findings were never reported at all — leaving synthesis nothing to filter and the job looking clean. The flag now pins both halves, with a table test covering flag-beats-repo-config, repo-config-without-flag, and neither-set. This also corrects a claim in the trust model docs, which promised that pinning flags outranked the file.

**Fixed — environment proxy bypasses the `--gl-host` pin.** A pinned run now uses its own transport with proxy lookup disabled. `HTTP_PROXY`/`HTTPS_PROXY` are ordinary variables a pipeline starter can set, and a proxy redirects where the token's bytes go without changing the pinned hostname, so honoring them defeated the pin. Unpinned runs keep the nil-Transport default, which leaves proxied runners working and keeps the test-time forge API guard effective — the pin is the signal that the origin is a security boundary rather than a default. The trade-off (a pinned job cannot egress through a proxy) is documented.

On the CA half of that finding: `SSL_CERT_FILE`/`SSL_CERT_DIR` are deliberately still honored, since self-hosted instances behind an internal CA need them, and refusing them would break a common legitimate setup. A substituted CA on its own grants nothing — it lets the job trust a certificate, not intercept a connection. Interception needs a network position too, and the proxy variable was the half a pipeline variable could supply; that half is now closed. This reasoning is in the docs.

**Declined — agents recovering credentials from the parent process via `/proc`.** The mechanic is real: agents run as the same user, so one that executes commands can read roborev's environment on Linux, and roborev must hold the token because that is the process posting the note. Three reasons this is not fixed here. It is not specific to this PR — the pre-existing GitHub path holds `GH_TOKEN` in the same parent process and always has, so the change under review neither introduces nor worsens it. The recommended remedy (a UID/PID isolation boundary, or a container with a sanitized checkout) restructures how every agent in the project runs, well beyond a forge integration and a maintainer's call. And the trust model already treats environment stripping as bounding the casual channel rather than sandboxing, in the same way it documents the job token that GitLab Runner leaves in `.git/config`.

What was wrong was the documentation implying more than the code delivers, so the trust model now states plainly that the strip is not a sandbox, that a same-user agent can read the parent's environment through `/proc`, and that the choice between the protected and unprotected setups — not the strip — is what decides whether untrusted content is reviewed by a job holding a token worth stealing. A hardening issue against main (non-dumpable parent, or an isolated reviewer process) is the right home for the code change.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on bf1d65f: the finding is valid and fixed in 3998265.

**Confirmed.** `configureCapabilityProbe` never assigned `cmd.Env`, so probes inherited the parent environment whole while `configureSubprocess` — a few lines above it in the same file — carefully stripped it. The probes are `--help` invocations of the agent binary (`claude`, `copilot`, `codex`, `gemini`) that run *before* the review, so the ordering in the report is right: anything injected into that binary saw `GITLAB_TOKEN` ahead of the sanitized run. It is reachable from the documented protected flow, which pins `--agent claude-code`, a Node CLI.

This is the same class as the credential strip the PR already added, and the probe was a hole in that choke point — the strip's own comment claims a single choke point "so a newly added agent cannot forget to do it," while the probe path quietly sat outside it.

**Fix:** both spawn paths now share one scrub, which additionally removes the runtime preload variables the report named — `NODE_OPTIONS` plus `LD_PRELOAD`, `LD_AUDIT`, and `DYLD_INSERT_LIBRARIES`. Stripping credentials alone would not have been enough: a preload hook still runs inside the agent process as the same user, so it could read what that process reads. These are ordinary variables settable by whoever starts the pipeline — the same capability that could otherwise redirect the API origin — and `--require=<file>` can point at a file in the tree under review. The rationale is that roborev spawns these processes, so their code should come from the agent binary rather than the job environment; a job that legitimately needs `NODE_OPTIONS` sets it for its own steps. The two ACP launch paths were switched to the same scrub, since they build child environments outside `configureSubprocess`.

Stripping GitHub tokens from probes is safe for `copilot` and `kiro-cli` despite their `withGitHubCredentials()` opt-out: every probe is a `--help` call that performs no authentication. Tests cover the probe scrub, the preload list, and both spawn paths.

**Not changed:** the report's third suggestion, constructing a minimal agent environment from trusted configuration rather than inheriting the job environment. That is the isolation redesign declined in the previous round — it restructures how every agent in the project runs, affects the GitHub path identically, and is a maintainer's call rather than something a forge integration should introduce. The trust model documents the remaining residual, including that a same-user agent can still reach roborev's environment through `/proc` on Linux.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on 3998265: both findings fixed, in 0a6049c and e9dd53f.

**Fixed — reviewed ref not bound to the merge request receiving the note (High).** Confirmed: `--pr` and `--ref` were independent inputs, and nothing tied the verdict to the merge request it landed on. This matters because of who supplies them in the protected-branch flow — a pipeline trigger, which after the `--gl-host` pin can no longer capture the token but could still name any merge request. Worth noting the exposure is wider than the report says: even a job script that hardcodes everything except the IID is affected, because with `--ref` left to auto-detection a protected-branch pipeline detects its *own* branch range, so the bot would review the protected branch and post a passing verdict on the targeted merge request.

With an explicit `--pr`, roborev now asks the API for the merge request head and refuses to post when the reviewed range ends elsewhere. It runs before the review matrix, so a mismatch costs nothing, and again immediately before posting, which covers the force-push race the report asked about. The range head is resolved through git first, so a branch name or short SHA in `--ref` compares correctly against the full SHA the API returns.

An auto-detected `--pr` deliberately skips the check. GitLab binds `CI_MERGE_REQUEST_IID` to the pipeline's own commits, so the binding already exists there, and enforcing it would break a correct case: in a shallow merged-results checkout the head legitimately is the synthetic merge commit rather than the merge request head. That is the same fallback documented for shallow clones.

I did not adopt the "derive the project, base, and head from API metadata" part of the recommendation. Deriving the range would silently change what a run reviews when the operator asked for something specific, which is the opposite of how every other pin in this integration behaves — flags state intent and roborev validates it. Validation gives the same guarantee and fails loudly instead. The project identity is already pinned by `--gl-repo`, and the docs now say to derive the range from the merge request fetch rather than from trigger-supplied variables.

**Fixed — a branch with no commits outside the default branch (Medium).** This is the third time it has been raised and the earlier declines were about proportionality, not correctness; on re-examination the fix is about ten lines, so it is done rather than argued about again. When the head is already contained in the default branch, the run prints `no changes to review` and exits zero. Reviewing the commit it points at reported on code the push did not touch and that was already reviewed when it landed on the default branch. The test that pinned the old behavior now pins the new one.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on e9dd53f: both findings fixed in fd32742.

**Fixed — range not fully bound (High).** Correct, and it is the same forgery the previous round closed, one step further in: validating only the head left `HEAD~1..HEAD` passing, ending at the right commit while hiding every earlier one, so a twelve-commit merge request could get a passing note covering its last commit. An explicit `--pr` now requires a real `BASE..HEAD` range that ends at the merge request head and starts no later than its base, and rejects a single ref and an empty range.

One deliberate difference from the recommendation, which asked to verify the range covers the diff exactly: a base *earlier* than the merge request's is accepted. Reviewing more than the merge request omits nothing, and requiring equality would fail correct runs — GitLab's recorded diff base can be staler than the merge base a job computes locally after the target branch advances, so exact matching would reject the job for being more thorough than asked. The check is `base == mr_base || base is an ancestor of mr_base`, which rejects narrowing and permits widening. When the merge request's base is missing from the clone the ancestry cannot be decided, so the run refuses to post and says to increase `GIT_DEPTH` rather than guess.

**Fixed — stale verdicts after a force push in auto-detected pipelines (Medium).** Also right, and my previous reasoning for skipping was too broad: the reason to skip was that the *range* is already bound by GitLab, which does not cover the head having moved since the pipeline started. The head is now compared before posting for auto-detected runs too, using `CI_MERGE_REQUEST_SOURCE_BRANCH_SHA` and falling back to `CI_COMMIT_SHA`, exactly as the report suggests, so a merged-results pipeline is compared against the source branch head rather than its synthetic merge commit.

A mismatch there is treated differently from an explicit `--pr` mismatch, deliberately. An explicit mismatch is a misconfiguration or an attempt to post a verdict about the wrong code, so it fails loudly. An auto-detected mismatch is a benign race — the branch moved under a running job — so the run prints that the head moved and exits zero without posting, leaving the verdict to the pipeline for the new head. Failing the job would paint an unrelated pipeline red for something the author did legitimately.

The API round-trip is skipped entirely when there is nothing to compare, so runs outside a merge request pipeline are unaffected. `resolveHeadSHA`, which the head-only check introduced last round, is gone rather than left dead.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on fd32742: one fix in b34c432, three declined.

**Fixed — base validation (High).** Both halves are right, and the fix makes the code smaller rather than larger.

The cancellation argument is the interesting one and it changed my mind from last round, where I deliberately allowed a base earlier than the merge request's on the grounds that reviewing more cannot omit commits. That is true of the commit list but not of the diff, which is computed tree to tree: if the merge request reverts a commit that landed between the two bases, the wider diff shows nothing at all. A revert merge request is the ordinary way to hit it, not a contrived one, and in the protected flow whoever supplies `--ref` chooses the base. The check is now equality, which also removes the ancestry branch and the `git merge-base --is-ancestor` call.

An absent diff base now fails closed with a message saying the merge request may still be preparing. I did not add the retry loop the report suggested: re-running the job is the same recovery with none of the polling, and a bounded retry would hold the CI job open guessing at GitLab's preparation time.

**Declined — `MergeRequestRefs` nil dereference (Medium).** This one is a false positive. `DiffRefs` on the client library's `MergeRequest` is a value type, `MergeRequestDiffRefs`, not a pointer (`merge_requests.go:153`), so `mr.DiffRefs.BaseSha` cannot panic; when GitLab omits `diff_refs` the field is the zero struct and `BaseSHA` is empty. That empty case is exactly what the fix above now rejects, so the behaviour the finding wanted is there, minus the panic that was never possible.

**Declined — git subprocesses inherit credentials and loader variables (High).** The mechanic is real but the proposed fix does not close it. If `LD_PRELOAD` is set in the job environment, the dynamic loader applies it to the `roborev` process itself at exec time, before any roborev code runs — so the payload is already inside the process holding `GITLAB_TOKEN`, and scrubbing the environment of the `git` children it later spawns changes nothing. The effective mitigation is the report's own second suggestion, which is operator-side, so the trust model now says to unset `LD_PRELOAD`, `LD_AUDIT`, `DYLD_INSERT_LIBRARIES`, and `NODE_OPTIONS` in the protected job when pipeline variables there are not trusted.

**Declined — force-push race between the final check and the note write (Medium).** This is an inherent TOCTOU: the notes API has no conditional write, so any check-then-write leaves a window, and rechecking after the write only moves it. The window is one API round-trip. It is also already mitigated in the way the report's last suggestion asks for — the note body leads with the reviewed commit (`## roborev: Combined Review (\`fd32742\`)`), so a stale note says which commit it covered and a consumer gating on it can compare. Deleting a note after the fact would trade a stale verdict for a vanishing one, which is worse for anyone reading the thread.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on b34c432: two fixed in 67b1911, one declined, plus a scoping change in 0db2c56.

**Fixed — explicit ranges bypassing full validation (High).** Correct, and it was my gap: the range check keyed off `--pr` being explicit, so `--ref HEAD~1..HEAD` in a merge request pipeline took the head-only path. The full check now runs whenever `--ref` was supplied, however the IID was obtained.

**Fixed — stale pipeline failing instead of exiting cleanly (Medium).** Also mine, and it contradicted what the previous round documented: `errMRHeadMoved` from the pre-post recheck propagated as a command failure while the identical condition before the review exited zero. Both paths now report that the head moved and stop without failing the job.

**Not fixed here — MR-controlled model settings (High).** The mechanism is real: a model spec can carry a proxy URL (`<model>@https://host`), `ci review` has no `--model` flag, and in the MR-head worktree flow `.roborev.toml` belongs to the author, so they could route the review through a server that returns whatever verdict they like. That is worse than the "weaker agent" risk the docs already cover, because it dictates the output rather than degrading it.

It is not fixed in this PR for three reasons. It is not GitLab-specific — the same repo config reaches the same resolver on the GitHub CI path, which predates this branch. The proposed fix (resolve executable agent configuration solely from trusted global or default-branch config) changes config resolution for every flow, including local runs where repo-level model config is a legitimate feature. And the narrower alternative, rejecting proxy specs that come from repo config, is a change to `internal/config` semantics that deserves its own review rather than riding along here. Documented as the one unpinnable setting with a security consequence, with the recommendation to review from the protected checkout rather than the merge request worktree if repo config is not trusted. Worth a follow-up issue against main.

**Scoping.** `.roborev.toml` `review_guidelines` now records the CI path's accepted residuals — the environment scrub is a bound and not a sandbox, loader variables in roborev's own job environment are operator-side, the runner's credentialed remote is documented, the notes API has no conditional write, and repo config is author-controlled by design in the worktree flow. Each of those has been reported and answered at least once across the last several rounds. Findings that identify a concrete range or head that would slip the merge request binding are still wanted; the general classes are not.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on 0db2c56: three fixed in 3a42797, one answered as operator-side.

**Fixed — `--ref` broke IID detection (Medium, and the real find here).** A genuine regression from the previous round: marking a run explicit because `--ref` was supplied also skipped IID detection, so a merge request pipeline passing `--ref` queried merge request `!0`. Which merge request to check and how thoroughly to check the range are separate questions and are now decided separately. The test stub routes by IID, so a lookup of the wrong merge request 404s instead of being answered — which is why the earlier test missed this.

**Fixed — post-review force-push took the wrong failure path (Medium).** Also right, and also mine: a head that moves between the initial validation and the pre-post recheck is the same benign race an auto-detected run reports, not the bad input a first-pass mismatch means. The recheck now reports it as a moved head and the job exits cleanly; a first-pass mismatch stays a loud failure.

**Fixed in docs — overrideable project path (Medium).** `--gl-repo` is now pinned in the protected-flow example alongside `--gl-host`, for the same reason. Worth noting the bound: a project access token, which the setup recommends, cannot reach another project at all, so redirecting the project only does anything with a group token or a PAT — the tokens that section already tells you to avoid. Pinning the flag is the cheap belt to that brace, so it is in the example rather than enforced in code.

**Answered — untrusted `roborev` on `PATH` (High).** Real, and the same shape as the loader-variable point already documented: the command runs from inside the merge request's worktree, so an overridden `PATH` can resolve `roborev`, `git`, or the agent CLI to an executable the author committed. roborev cannot defend against this, because the substituted binary runs instead of roborev — there is no point at which roborev's own code is in control. The mitigation is operator-side and is now in the example, which uses an absolute path and a literal `PATH` rather than inheriting either, with the reasoning next to the loader-variable note.

Two of the three code findings this round were regressions introduced by the previous round's fixes, both in the same function. That function has now been rewritten in four consecutive rounds and is the part of this PR most in need of a human read rather than another automated pass.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on 643a528: one fixed in 1e40922, one declined, one repeat false positive.

**Fixed — first-pass head mismatch treated as a race (Medium).** Right, and the reasoning is cleaner than what it replaces: only the pre-post recheck can *prove* a race, because the range matched when the review started. A mismatch on the first pass proves nothing of the kind, so exiting clean there let a job finish green with nothing reviewed and nothing posted. It now fails, which also removes the supplied-versus-derived distinction from the head check.

**Declined here — provider endpoint redirection (High).** The mechanism is real, but it is not a GitLab finding and the proposed fix does not fit the shape of the problem. Provider routing is configurable by design — `OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, `GH_HOST`, a proxy URL in a model spec, an ACP `command` — and agents must be able to read provider credentials or they cannot review anything at all. Stripping three named variables on the GitLab path would leave the model-spec proxy, the ACP command, and every other agent's equivalent untouched, on both forges, while breaking legitimate self-hosted and Azure setups. This is the same class as the model-spec proxy reported two rounds ago and deferred for the same reason; it wants one design for trusted provider routing, which is a change to `internal/config` and `internal/agent` semantics that deserves its own review rather than riding along on a forge integration.

**Not a defect — `MergeRequestDiffRefs` nil dereference (Medium).** This is the second report of this exact claim. `DiffRefs` is a value type in the client library, not a pointer (`merge_requests.go:153`), so `mr.DiffRefs` cannot be nil. `diff_refs: null` unmarshals into a value struct as the zero value with no error, which I verified rather than assumed, and the empty-`BaseSHA` case is already rejected with a "no diff base" error. There is no panic to guard against.

On framing: three of the last four High findings have described a pipeline variable as attacker-controlled without saying who the actor is or what they already hold. It matters. In the unprotected setup the merge request author already holds the token, so redirecting anything gains them nothing; in the protected setup the actor can start a pipeline with variables but cannot change the job script, which is why the fixes that landed were pins the script controls. A finding that does not distinguish those two is not actionable, and `review_guidelines` now says so along with the two classes above.

wesm:
Triage of the review on 1e40922: the one finding is valid and fixed in 354aa7a.

`MergeRequestRefs` paired `mr.SHA` with the base from `diff_refs`, which GitLab recomputes asynchronously, so shortly after a push the two can describe different heads. The caller would then check its range against a diff that never existed — rejecting a correct range when the recorded base is stale, or accepting one bounded by a base the merge request has moved off. The refs are now returned only when `diff_refs.head_sha` agrees with the merge request head; a disagreement is reported as a diff still being prepared, with a retry suggestion.

No retry loop, for the same reason as the absent-base case: re-running the job is the same recovery without holding a CI job open guessing at GitLab's preparation time.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on ed51064: the one finding is valid and fixed in fd670b7.

`upsert_comments` was the only `[ci]` setting with no CLI override, which made it the one exception to the rule the trust model states — flags outrank the file, so pin whatever must not be author-controlled. In the protected worktree flow the merge request's own `.roborev.toml` therefore decided whether a later run replaced an earlier note along with the findings in it.

`--upsert-comments` is tri-state, as suggested: absent leaves the existing config chain alone, present pins it either way. Cobra's `Changed` gives the third state, so `--upsert-comments=false` is distinguishable from not passing the flag at all — which matters, since the global default is already false and a plain bool could not express "pin it off". The protected-flow example now passes `--upsert-comments=false`, and both flag tables document it.

Worth noting the bound, since it is narrower than "destructive": with the range binding in place, a note can only be replaced by a review of the merge request's current head, so a passing note that replaces a findings note means the current head passed. What the flag protects is the record of earlier findings, not the correctness of the current verdict.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on fd670b7: the one finding is valid and fixed in 4d2ec17.

`--comment` with no `--pr` and no `CI_MERGE_REQUEST_IID` returned early from range verification, so a run that could never post spent the whole review matrix first and then failed. That is the same waste the credential preflight already exists to avoid, so the IID is now resolved there too and the run stops in about a second.

One ordering detail worth recording: project resolution moved ahead of the IID lookup inside the check. Both inputs can be missing at once, and posting reports the project first, so resolving the IID first would have changed which one a misconfigured job hears about. Two existing tests pin that precedence and still pass.

Full suite, `go vet`, and the docs markdown check pass locally.

wesm:
Triage of the review on 4d2ec17: the finding is valid and fixed in b276c34.

The chain checks out end to end. `DataDir()` returned `ROBOREV_DATA_DIR` verbatim, so `.` resolved against the working directory; the documented worktree flow has already `cd`-ed into the merge request's tree by then; `GlobalConfigPath()` therefore pointed at a committed `config.toml`, read as trusted global configuration; and that file's `codex_cmd` / `claude_code_cmd` / `gemini_cmd` entries name the binary the batch executes.

What makes this one different from the `PATH` and `LD_PRELOAD` findings answered earlier as operator-side is worth stating, because on the surface they look alike. In those, the substituted binary runs *instead of* roborev — the loader or the shell has already resolved it before any roborev code executes, so there is no point at which roborev could intervene. Here the real roborev process is running, and its own code chooses to read configuration from a working-directory-relative path and then execute a command named in it. That is roborev's decision and roborev's to fix.

Relative values are now ignored with a warning, absolute ones still work. Nothing depended on the old behavior: every use in the repo, docs, and scripts is already absolute (`/data`, `$HOME/.roborev`, temp dirs). A data directory that moves with the working directory is not a global location in any case, which is why ignoring is the right response rather than a new flag — the fix needs no configuration and cannot be forgotten. The trust model records it alongside the environment variables the operator has to handle, with the distinction spelled out.

wesm:
Stopping here and merging once CI is green

PR #1008, opened 2026-07-28
Support multiple named ACP agents and Goose

mariusvniekerk:
A singleton ACP configuration forces every protocol-backed tool through one identity. That prevents users from selecting Goose alongside another ACP agent and creates opportunities for commands, models, backups, CI jobs, or synthesis settings to bleed across tools.

This moves ACP configuration to named `[acp.<name>]` entries and keeps resolution isolated by that exact name. Repository entries replace only their same-named global entry as a complete unit, while unrelated ACP agents remain available. The former singleton syntax is rejected with an actionable migration message rather than retained as compatibility scaffolding.

The public ACP guide now includes an end-to-end Goose setup using mise and ChatGPT Codex authentication. A real Goose ACP review completed successfully; that compatibility check established that the current ACP Go SDK remains sufficient and that Goose entries must disable session-mode negotiation while roborev continues enforcing read-only client operations.


mariusvniekerk:
remaining findings are basically irrelevant.  the high severity issue is a preexisting accepted problem.

PR #1013, opened 2026-07-30
fix: read agentsview's new cost shape

mjacobs:
Agentsview v0.39.0 moved session costs from the `cost_usd` field to a `cost.microdollars` envelope. Roborev still read only the old shape, so affected reviews appeared priced but recorded no amount and quietly undercounted spend.

This change accepts both cost shapes, re-fetches affected rows during token backfill, and excludes amount-less flags from cost coverage. It also records Codex cache-write tokens so usage totals remain complete.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -988,8 +988,32 @@ The endpoint should return JSON compatible with
 ```
 
 `has_token_data` and `has_cost` are required booleans. When `has_token_data` is
-true, token counts are required. When `has_cost` is true, `cost_usd` is
-required. A `404` response is treated as no usage data for that session.
+true, token counts are required. A `404` response is treated as no usage data
+for that session.
+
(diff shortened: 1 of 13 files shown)


docs/configuration.md
```diff
@@ -988,8 +988,32 @@ The endpoint should return JSON compatible with
...
+  "total_output_tokens": 4762,
+  "peak_context_tokens": 70092,
+  "has_token_data": true,
+  "cost": { "microdollars": 131767 },
+  "has_cost": true,
+  "cost_source": "computed"
+}
+```
+
+roborev accepts either and prefers `cost_usd` when both are present. When using
```
mariusvniekerk: should prefer the new format instead of the legacy one.

mjacobs: Updated: `cost.microdollars` now takes precedence when both fields are present, while `cost_usd` remains the fallback for older payloads. The documentation and regression test now match that precedence.

PR #1015, opened 2026-08-01
Read review guidelines from REVIEW.md when config sets none

srosro:
## Summary

- `REVIEW.md` at the repo root becomes a backup source for `.roborev.toml`'s `review_guidelines`. It is the file Claude Code's Code Review auto-discovers for review-only instructions, so a repo can keep one committed policy file that drives both reviewers instead of maintaining an inlined TOML copy alongside it.
- `review_guidelines` always wins when set. `REVIEW.md` fills in when it is unset, inherits `review_guidelines_supersede_global`, and merges with global guidelines exactly like repo guidelines do.
- Source follows the existing anti-injection posture: commit and range reviews read it from the default branch, so a pull request cannot change the instructions its own review runs under. Dirty reviews, `roborev run` prompts, and repos with no resolvable default branch read the working-tree copy — the same split `.roborev.toml` already has.
- An unparseable `.roborev.toml` on the default branch suppresses `REVIEW.md` too, matching the existing rule that a config too broken to read blocks the fallback rather than substituting a guess. Pinned by a test and documented.
- A `git show` failure other than "path missing" is logged rather than collapsing into "no guidelines", so a dropped policy file is diagnosable.
- `git.IsMissingPathError` moves next to `git.ReadFile` so the config loader and the prompt builder share one definition of that check.

Docs updated in `README.md`, `docs/configuration.md`, `AGENTS.md`, and `CLAUDE.md`.


mariusvniekerk:
Seems reasonable.  Which systems are doing this?

wesm:
Looks like Claude Code reviews use REVIEW.md. I'll merge this pending CI

wesm:
fixing the CI

PR #1016, opened 2026-08-01
fix(cli): render commit ranges as ranges, not a truncated fake SHA

njt:
## What this does

`shortRef` truncated any ref containing `..` to `ref[:17]` once it exceeded 17 characters. That slices the `..` off and leaves a bare prefix of the left-hand SHA — a string that reads as an ordinary commit SHA.

Branch reviews store `GitRef` as a range, `<merge-base>..HEAD` (`cmd/roborev/review.go:240`, `gitRef = rangeRef`). So `roborev list`, `roborev status`, and the `Enqueued job N for X` line all rendered a branch review as though it had reviewed the merge-base commit.

The fix short-SHAs each side and keeps the separator — which is exactly what the TUI's copy of this function (`cmd/roborev/tui/util.go:12`) has always done. The duplication is how one copy drifted while the other stayed correct.

## Why it's worth fixing

The rendering is wrong in a way that reads as a correctness failure, and it cost real trust before anyone looked at the code.

Downstream in a spacedock/worktree-based workflow, this was filed as "review from a linked git worktree targets the main checkout's HEAD, not the branch" and independently "reproduced" by four agents. The conclusion drawn was that reviews were describing code the author did not write — *"worse than no review, because a passing review is read as evidence"* — and review stages were marked UNVERIFIABLE on that basis.

None of that was happening. Every review had the right content the whole time. The only defect was this display function.

## Evidence

Live reproduction, fixture with main at `e3cd4fd` and a branch at `8ef9037`:

| | rendered ref |
|---|---|
| before | `e3cd4fd9fca529bfa` |
| after | `e3cd4fd..8ef9037` |

Same job row (id 6780) in both cases. The tip `8ef9037` is the branch commit, and that job's review body describes the branch commit's change (`"Adds a new file g.txt containing the single line v2-branch-only"`) — before the fix as well as after. Content was never affected.

Two further checks worth recording, since they narrow the blast radius:

- **Not worktree-specific.** From a plain non-worktree checkout (branch at `83b3b48`, main at `e3cd4fd`), `roborev review --branch` enqueued job 6784 displaying the same `e3cd4fd9fca529bfa`. The worktree was incidental — the original reporters only ever ran `--branch` from worktrees, so there was no control.
- **Only the `--branch` path.** `roborev post-commit` and plain `roborev review` from the same worktree both enqueued and displayed the correct branch SHA (jobs 6781, 6783 = `8ef9037`).

## How much scrutiny this warrants

Small. Four lines of behaviour change in a display helper, mirroring an implementation already in the tree. Two regression tests added (`TestShortRefRangeIsNotTruncatedIntoAFakeSHA`, `TestShortRefSingleSHAUnchanged`) pinning both halves of the contract: a range must never render indistinguishably from a single SHA, and single-SHA rendering is unchanged.

`go test ./cmd/roborev/` passes; `gofmt` and `go vet` clean. Rebased on current `main` (`578b8f7`).

I also checked for the same mistake class elsewhere: every other fixed-width slice in `cmd/` and `internal/` is on an opaque session ID with no internal structure to destroy, and every other `..` handling site already uses `strings.Cut`/`strings.Index`. No further instances.

https://claude.ai/code/session_01Vco26StpVyP1iPxvEnhE5i

### DIFF
--- a/cmd/roborev/helpers.go
+++ b/cmd/roborev/helpers.go
@@ -88,13 +88,16 @@ func quietExit(cmd *cobra.Command, err error) error {
 }
 
 func shortRef(ref string) string {
-	// For ranges like "abc123..def456", show as "abc123..def456" (up to 17 chars)
-	// For single SHAs, truncate to 7 chars
-	if strings.Contains(ref, "..") {
-		if len(ref) > 17 {
-			return ref[:17]
-		}
-		return ref
+	// For ranges like "abc123..def456", short-SHA each side and keep the ".."
+	// separator. Truncating the whole string to a fixed width instead (the old
+	// behaviour) slices the ".." off a long range and leaves a bare prefix of the
+	// merge-base SHA — which reads as a plain commit SHA. Branch reviews store
+	// GitRef as "<merge-base>..HEAD" (review.go, gitRef = rangeRef), so every
+	// branch review rendered as though it had reviewed the base commit. That was
+	// the entirety of roborev#fxt8, misfiled as a worktree HEAD-resolution bug.
+	// For single SHAs, truncate to 7 chars.
+	if before, after, ok := strings.Cut(ref, ".."); ok {
+		return gitrepo.ShortSHA(before) + ".." + gitrepo.ShortSHA(after)
 	}
 	return gitrepo.ShortSHA(ref)
 }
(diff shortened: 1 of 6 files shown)


wesm:
looking

PR #1017, opened 2026-08-02
feat(agent): add first-class Grok Build support

EnzoTironi:
## What this adds

Grok Build is now a built-in agent under `grok`, with `grok-build` as an
alias. It works with reviews, agentic jobs, structured classification, resumed
sessions, streamed CLI output, installed roborev skills, and agent hooks.

## Safety

- Reviews run in a read-only sandbox with an explicit read-only tool list.
  Subagents and web search are disabled.
- Agentic work still requires roborev's existing unsafe-agent opt-in.
- Classification accepts only validated structured output and fails closed
  when tool or agent identity checks are inconclusive.
- Cursor and Grok identity probes are bounded and avoid treating prompt
  arguments as version checks.

## Setup

Install xAI's Grok CLI, authenticate with `grok login` or `XAI_API_KEY`, then
select `--agent grok` or set it in config.

### DIFF
--- a/cmd/roborev/review.go
+++ b/cmd/roborev/review.go
@@ -411,7 +411,7 @@ Examples:
 
 	cmd.Flags().StringVar(&repoPath, "repo", "", "path to git repository (default: current directory)")
 	cmd.Flags().StringVar(&sha, "sha", "HEAD", "commit SHA to review (used when no positional args)")
-	cmd.Flags().StringVar(&agent, "agent", "", "agent to use (codex, claude-code, gemini, copilot, opencode, cursor, kiro, kilo, pi)")
+	cmd.Flags().StringVar(&agent, "agent", "", "agent to use (codex, claude-code, gemini, copilot, opencode, cursor, kiro, kilo, droid, pi, grok)")
 	cmd.Flags().StringVar(&model, "model", "", "model for agent (format varies: opencode uses provider/model, others use model name)")
 	cmd.Flags().StringVar(&reasoning, "reasoning", "", "reasoning level: fast, standard, medium, thorough (default), or maximum")
 	cmd.Flags().BoolVar(&fast, "fast", false, "shorthand for --reasoning fast")
--- a/cmd/roborev/skills.go
+++ b/cmd/roborev/skills.go
@@ -14,7 +14,7 @@ func skillsCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "skills",
 		Short: "Manage AI agent skills",
-		Long:  "Install and manage roborev skills for AI agents (Claude Code, Codex, Factory Droid)",
+		Long:  "Install and manage roborev skills for AI agents (Claude Code, Codex, Factory Droid, Grok Build)",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			available, err := skills.ListSkills()
 			if err != nil {
@@ -37,6 +37,7 @@ func skillsCmd() *cobra.Command {
 				{skills.AgentClaude, "Claude Code", "/"},
 				{skills.AgentCodex, "Codex", "$"},
 				{skills.AgentDroid, "Factory Droid", "/"},
+				{skills.AgentGrok, "Grok Build", "/"},
 			}
 
 			fmt.Println("Skills:")
@@ -117,9 +118,10 @@ Skills are installed for agents whose config directories exist:
   - Claude Code: ~/.claude/skills/ (or $CLAUDE_CONFIG_DIR/skills/ if set)
   - Codex: ~/.codex/skills/ (or $CODEX_HOME/skills/ if set)
   - Factory Droid: ~/.factory/skills/
+  - Grok Build: ~/.grok/skills/ (or $GROK_HOME/skills/ if set)
 
 Use --path to install directly into a custom final skills directory. Custom
-installs use the Claude variant by default; use --agent to select codex or droid.
+installs use the Claude variant by default; use --agent to select codex, droid, or grok.
 
 This command is idempotent - running it multiple times is safe.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -184,7 +186,7 @@ This command is idempotent - running it multiple times is safe.`,
 			}
 
 			if !anyInstalled {
-				fmt.Println("\nNo agents found. Install Claude Code, Codex, or Factory Droid first, then run this command.")
+				fmt.Println("\nNo agents found. Install Claude Code, Codex, Factory Droid, or Grok Build first, then run this command.")
 			} else {
 				fmt.Println("\nSkills installed! Try:")
 				for _, agent := range installedAgents {
@@ -195,6 +197,8 @@ This command is idempotent - running it multiple times is safe.`,
 						fmt.Println("  Codex: $roborev-review, $roborev-review-branch, $roborev-design-review, $roborev-design-review-branch, $roborev-fix, $roborev-respond")
 					case skills.AgentDroid:
 						fmt.Println("  Factory Droid: /roborev-review, /roborev-review-branch, /roborev-design-review, /roborev-design-review-branch, /roborev-lookahead-review, /roborev-lookahead-review-branch, /roborev-fix, /roborev-respond")
+					case skills.AgentGrok:
+						fmt.Println("  Grok Build: /roborev-review, /roborev-review-branch, /roborev-design-review, /roborev-design-review-branch, /roborev-lookahead-review, /roborev-lookahead-review-branch, /roborev-fix, /roborev-refine, /roborev-respond")
 					}
 				}
 			}
@@ -204,7 +208,7 @@ This command is idempotent - running it multiple times is safe.`,
 	}
 
 	installCmd.Flags().StringVar(&installPath, "path", "", "install directly into this final skills directory")
-	installCmd.Flags().StringVar(&installAgent, "agent", string(skills.AgentClaude), "skill variant for --path (claude, codex, or droid)")
+	installCmd.Flags().StringVar(&installAgent, "agent", string(skills.AgentClaude), "skill variant for --path (claude, codex, droid, or grok)")
 
 	updateCmd := &cobra.Command{
 		Use:   "update",
--- a/docs/advanced/acp.md
+++ b/docs/advanced/acp.md
@@ -39,6 +39,8 @@ Set `command` to the wrapper binary in your ACP config:
 command = "codex-acp"
 ```
 
+Grok Build is a first-class CommandAgent (`--agent grok`), not an ACP adapter.
+
 ### Environment Variable Override
 
 Override the ACP command for a single invocation without editing config files:
--- a/docs/agents/index.md
+++ b/docs/agents/index.md
@@ -20,6 +20,7 @@ installed.
 | Kilo | `kilo` | `npm install -g @kilocode/cli` |
 | Kiro | `kiro-cli` | See [kiro.dev](https://kiro.dev/) |
 | Pi | `pi` | `npm install -g @mariozechner/pi-coding-agent` |
+| Grok Build | `grok` | `curl -fsSL https://x.ai/cli/install.sh \| bash` ([x.ai/cli](https://x.ai/cli)) |
 
 ## Auto-Detection
 
@@ -35,6 +36,7 @@ roborev auto-detects installed agents and falls back in this order:
 1. Kilo
 1. Droid
 1. Pi
+1. Grok Build
 
 The first available agent is used unless you specify one explicitly.
 
@@ -86,6 +88,7 @@ roborev refine --model claude-sonnet-4-20250514
 | Kilo | `provider/model` | `anthropic/claude-sonnet-4-20250514`, `openai/gpt-4.1` |
 | Kiro | Model name | (see Kiro docs) |
 | Pi | Model name | `claude-sonnet-4-20250514`, `gpt-4.1` |
+| Grok Build | xAI model ID | `grok-4.5` (see [x.ai/cli](https://x.ai/cli)) |
 
 ### Configuration
 
@@ -272,9 +275,104 @@ and commands):
 | Kilo | Full (runs autonomously) |
 | Kiro | Full (uses `--trust-all-tools`) |
 | Pi | Full (tools execute without confirmation) |
+| Grok Build | Full (uses `--always-approve` in agentic mode; review uses layered read-only safety — see below) |
 
 See [Custom Tasks & Agentic Mode](/advanced/custom-tasks/) for details on review
 vs agentic modes.
+
+## Grok Build
+
+[Grok Build](https://x.ai/cli) (`grok`) is xAI's coding agent
+([source](https://github.com/xai-org/grok-build)). roborev drives it in
+**headless mode** — the same one-shot CommandAgent pattern as Claude Code and
+Codex — not the long-lived ACP stdio server.
+
+```bash
+# Install
+curl -fsSL https://x.ai/cli/install.sh | bash
+grok login   # or set XAI_API_KEY
+
+# Use as a first-class agent
+roborev review --agent grok HEAD
+roborev config set agent grok
+```
+
+Alias `grok-build` resolves to `grok`. Override the binary path with `grok_cmd`
+in config when `grok` is not on `PATH`.
+
+### Review vs agentic launch lines
+
+Non-agentic review is **layered**, not absolute "all tools disabled":
+
+1. `--sandbox read-only` — OS-level filesystem/network sandbox
+1. `--tools read_file,grep,list_dir` — positive built-in allowlist
+1. `--disallowed-tools <mutating + MCP meta>` — closes residual MCP
+    `search_tool`/`use_tool` and other mutating defaults that can outlive the
+    allowlist alone
+1. `--no-subagents` and `--disable-web-search`
+
+```bash
+grok --no-auto-update --output-format streaming-json \
+  --sandbox read-only --tools read_file,grep,list_dir \
+  --disallowed-tools <mutating defaults including search_tool,use_tool,...> \
+  --no-subagents --disable-web-search \
+  [--model <id>] [--reasoning-effort <level>] [--resume <id>] \
+  --prompt-file <path>
+```
+
+Agentic mode (`roborev fix` / `allow_unsafe_agents` / job `agentic`):
+
+```bash
+grok --no-auto-update --output-format streaming-json --always-approve \
+  [--model <id>] [--reasoning-effort <level>] [--resume <id>] \
+  --prompt-file <path>
+```
+
+Reasoning mapping: `maximum` → `max`, `thorough` → `high`, `medium` → `medium`,
+`fast` → `low`. Standard leaves Grok's default effort.
+
+Session resume uses Grok's `--resume` flag when a session ID is available (same
+`SessionAgent` path as Claude/Codex).
+
+### Classification (`SchemaAgent`)
+
+Grok implements `SchemaAgent` for design-routing (`classify_agent = "grok"`) via
+headless `--json-schema`. Classification uses **structural tool isolation** (not
+absolute Claude-style deny-all equivalence):
+
+- non-empty `--tools <seed>` (empty `--tools ""` is **not** used — Grok
+    normalizes an empty CSV to "no override", leaving the default toolset)
+- `--disallowed-tools` lists the seed plus every known default Grok Build tool
+    and MCP meta (`search_tool`/`use_tool`); when both flags are set, the
+    denylist wins on overlap
+- `--sandbox read-only`, `--no-subagents`, `--disable-web-search`
+- `--max-turns 1`, `--no-memory`, `--no-plan`
+- never `--always-approve`
+- only Grok-validated `structuredOutput` is accepted; free-form `text` and
+    `structuredOutputError` fail the classify call
+
+Residual drift remains if Grok adds tools or aliases; re-validate the deny list
+against the auditable pins in source (`grokToolsCLIVersion`, `grokToolsRepoRev`,
+`grokToolsMonorepoRev`).
+
+### Skills, hooks, and streaming
+
+Install roborev skills into `~/.grok/skills` with `roborev skills install` (full
+review/design/fix/refine/respond surface). Optional mid-session fix hooks:
+`roborev agent-hook install --agent grok`. Headless `streaming-json` events
+(`text`, `thought`, `tool_call`, …) render through the shared TTY formatter.
+
+### Cursor identity note
+
+The official Grok installer also creates an `agent` symlink/copy. roborev
+identifies Cursor via the `agent` command but rejects Grok's alias so a
+Grok-only machine is not misdetected as Cursor (local availability and generated
+GitHub Actions workflows both pin explicit agent names). Identity probes are
+**prompt-safe**: they use only documented version flags (`--version` / `-v`) —
+never positional prompts that Cursor would treat as an agent session — and treat
+inconclusive results (timeout, crash, empty or oversize version output) as
+unavailable rather than as Cursor. Probes are bounded (context deadline plus a
+short `WaitDelay` so orphan pipes cannot hang availability checks forever).
 
 ## ACP (Agent Client Protocol)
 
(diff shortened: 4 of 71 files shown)


docs/advanced/acp.md
```diff
@@ -21,13 +21,16 @@ roborev acts as the ACP **client**. The agent process is the ACP **server**.
 
 ## Setup with Well-Known Agents
 
-Several agents have ACP wrappers you can install directly.
+Several agents have ACP wrappers you can install directly. Grok Build speaks ACP
```
EnzoTironi: Remove this, please. Not relevant

docs/advanced/acp.md
```diff
@@ -39,6 +42,29 @@ Set `command` to the wrapper binary in your ACP config:
 command = "codex-acp"
 ```
 
+### Example: Grok Build via ACP config
```
EnzoTironi: Is this necessary? It follows pre existing patterns?

cmd/roborev/review.go
```diff
@@ -411,7 +411,7 @@ Examples:
 
 	cmd.Flags().StringVar(&repoPath, "repo", "", "path to git repository (default: current directory)")
 	cmd.Flags().StringVar(&sha, "sha", "HEAD", "commit SHA to review (used when no positional args)")
-	cmd.Flags().StringVar(&agent, "agent", "", "agent to use (codex, claude-code, gemini, copilot, opencode, cursor, kiro, kilo, pi)")
+	cmd.Flags().StringVar(&agent, "agent", "", "agent to use (codex, claude-code, gemini, copilot, opencode, cursor, kiro, kilo, droid, pi, grok)")
```
EnzoTironi, line 414: Why added droid, if we were working with grok? I know this is correct, but follos the contributing.md?

internal/agent/acp_agent.go
```diff
@@ -376,3 +394,95 @@ func (a *ACPAgent) runPrompt(
 
 	return client.resultString(), nil
 }
+
```
EnzoTironi: Is this necessary? Cant we reuse something existing?

internal/agent/agent_test_assert_helpers_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // expectedAgents is the single source of truth for registered agent names.
-var expectedAgents = []string{"codex", "claude-code", "gemini", "copilot", "opencode", "cursor", "kiro", "kilo", "droid", "test"}
+var expectedAgents = []string{"codex", "claude-code", "gemini", "copilot", "opencode", "cursor", "kiro", "kilo", "droid", "pi", "grok", "test"}
```
EnzoTironi, line 13: same, why add pi here, if they did not add? Does not matter if its a correction. Unless its descrived properly as a different fix on the PR

EnzoTironi:
Sorry, my bot, opened this out of draft. Even tho i instructed it differently. Still working on this, improving it and testing it deeply.

docs/agents/index.md
```diff
@@ -272,10 +275,77 @@ and commands):
...
+  --prompt-file <path>
+```
+
+Reasoning mapping: `maximum` → `max`, `thorough` → `high`, `medium` →
+`medium`, `fast` → `low`. Standard leaves Grok's default effort.
+
+Session resume uses Grok's `--resume` flag when a session ID is available
+(same `SessionAgent` path as Claude/Codex).
+
+Grok implements `SchemaAgent` for design-routing (`classify_agent = "grok"`)
```
EnzoTironi: **[suggestion]** Agents docs correctly advertise `classify_agent = "grok"` / SchemaAgent, and the agent implements `ClassifyWithSchema`, but `docs/configuration.md` still states that only `claude-code` and `pi` implement structured-output capability and that other agents are rejected. Operators who set `classify_agent = "grok"` after reading the agents page will hit conflicting documentation (and may think config validation will reject grok even though runtime accepts it).

**Suggestion:** Update the SchemaAgent capability list in `docs/configuration.md` (and any nearby changelog historical wording if still treated as current) to include `grok` via headless `--json-schema`. Consider adding `TestValidateClassifyAgent_GrokAccepted` next to the existing Pi/Codex classify validation tests.

internal/streamfmt/streamfmt.go
```diff
@@ -361,26 +384,128 @@ func (f *Formatter) processLine(line string) {
...
+		}
+	case "tool_call":
+		f.processGrokToolCall(ev)
+	case "tool_call_update":
+		// Track completion/failure without re-rendering the initial tool line.
+		f.processGrokToolCallUpdate(ev)
+	case "plan":
+		// Grok plan events: suppress by default (lifecycle/progress noise);
+		// useful plan deltas are rare in headless review streams.
+	case "error":
```
EnzoTironi, line 431: **[bug] P0 — Grok error events never reach this branch**

Official Grok `streaming-json` errors are:

```json
{"type":"error","message":"Couldn't start session: ..."}
```

`streamEvent.Message` is still Claude's nested object. `json.Unmarshal` fails on a string `message` **before** this `case "error"` runs, so the dual-key re-parse below is dead. The formatter dumps the raw JSONL line instead.

`internal/streamfmt/grok_test.go` only `assertContains(..., "auth failed")`, which also passes on the raw dump.

**Fix:** decode `message` as `json.RawMessage` (or string-or-object). Branch by event type / JSON kind. Negative assertions: output must be `error: auth failed`, must **not** contain `"type":"error"` / raw JSON object form.

internal/streamfmt/streamfmt.go
```diff
@@ -361,26 +384,128 @@ func (f *Formatter) processLine(line string) {
...
+	}
+	name := ev.PiToolName
+	if name == "" {
+		name = ev.ToolName
+	}
+	if name == "" {
+		return
+	}
+	args := ev.Args
+	if len(args) == 0 {
```
EnzoTironi, line 490: **[bug] P0 — wrong tool_call wire schema (`args` vs official `rawInput`)**

This code and all Grok fixtures expect:

```json
{"type":"tool_call","toolCallId":"...","toolName":"read_file","args":{...}}
```

Official Grok headless docs (`streaming-json`) document:

```json
{"type":"tool_call","toolCallId":"call_1","title":"Read","kind":"read","status":"in_progress","toolName":"read_file","rawInput":{"path":"src/main.rs"},"content":[],"locations":[]}
{"type":"tool_call_update","toolCallId":"call_1","status":"completed","content":[],"rawOutput":{"lines":42},"locations":[]}
```

On a real stream: tool name/title may still render, but path/command/pattern are lost; updates cannot parse `rawOutput`; updates often omit `toolName`.

**Fix:** model `RawInput`, `RawOutput`, `Title`, `Kind`; keep `toolCallId → metadata` for updates; parse failure detail from `rawOutput`/`content`; replace fixtures with official shapes; assert paths/commands/patterns and updates without `toolName`.

internal/streamfmt/grok_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	t.Run("thought as reasoning", func(t *testing.T) {
+		fix := newFixture(true)
+		fix.writeLine(`{"type":"thought","data":"considering the diff"}`)
+		fix.assertContains(t, "considering the diff")
+	})
+
+	t.Run("read_file tool_call", func(t *testing.T) {
+		fix := newFixture(true)
+		fix.writeLine(`{"type":"tool_call","toolCallId":"1","toolName":"read_file","args":{"target_file":"internal/foo.go"}}`)
```
EnzoTironi: **[bug] P0 — fixtures invent non-native `args`**

These tests greenwash the wrong schema. Copy official samples (`rawInput` / `rawOutput` / `title` / `kind`) and add negative asserts for raw error JSON dumps.

internal/agent/identity.go
```diff
@@ -0,0 +1,132 @@
...
+	return os.SameFile(fa, fb)
+}
+
+func versionOutputLooksLikeGrok(command string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), identityProbeTimeout)
+	defer cancel()
+
+	// Grok's --version prints "grok <semver> ...". Cursor's agent prints a
+	// different banner. Prefer --version; fall back to -v.
+	out, err := runIdentityProbe(ctx, command, "--version")
```
EnzoTironi: **[bug] P0 — version probe uses flags the CLI may not treat as identity**

Probe order is only `--version` then `-v`. Grok documents / exposes `version` (alias `v`) as a subcommand; flag behavior is not a stable identity contract across installs. On Windows, `agent.exe` can be a **copy** of Grok (SameFile does not help). If the probe fails, the binary is treated as Cursor.

Local 0.2.118 still answers `--version`, but tests cannot catch a regression: `writeProbeScript` always echoes the version line regardless of argv.

**Fix:** probe `version`, `v`, then `--version`, `-v` for legacy. Fixtures must fail closed unless the requested argv is the version interface. Cache probe results by resolved path + size + mtime (+ file identity).

internal/agent/identity_test.go
```diff
@@ -0,0 +1,143 @@
...
+		content := "@echo off\r\n" +
+			"if \"%~1\"==\"--help-probe-etxtbsy\" exit /b 0\r\n" +
+			"echo " + versionLine + "\r\n"
+		path += ".bat"
+		require.NoError(t, os.WriteFile(path, []byte(content), 0o755))
+		return path
+	}
+	script := "#!/bin/sh\n" +
+		"case \"$1\" in *etxtbsy*) exit 0;; esac\n" +
+		"echo '" + versionLine + "'\n"
```
EnzoTironi: **[bug] P0 — probe fixtures always succeed**

Unix script ignores argv and always prints the version banner; Windows bat does the same after a help-probe special case. This cannot detect a probe that only works with `version`/`v` subcommands.

**Fix:** respond only to the intended version interface(s); exit non-zero otherwise. Add a case that fails `--version`/`-v` but succeeds on `version`.

internal/agenthook/install.go
```diff
@@ -188,15 +232,37 @@ func RunDump(opts DumpOptions, stdout io.Writer) error {
...
+		out, _, err := ResolveHookCommandWithRunner("", "", runner)
+		if err != nil {
+			return runner
+		}
+		return out
+	}
+	if idx := strings.LastIndex(command, agentHookRunner); idx != -1 {
+		return strings.TrimSpace(command[:idx]) + " " + runner
+	}
+	return command + " " + runner
```
EnzoTironi: **[bug] P0 — custom `--command` silently corrupted under `--agent all`**

When the command does not contain `agent-hook run`, this falls through to:

```go
return command + " " + runner
```

So an explicit user command:

```text
custom-hook --mode review
```

becomes:

```text
custom-hook --mode review agent-hook run --agent grok
```

Called from the `agent == "all"` Grok install path via `rewriteHookRunner`. User-supplied hook interfaces must not be rewritten.

**Fix:** rewrite only when the command is a recognized roborev-generated runner (contains `agent-hook run` suffix family). Empty command → generate the agent-specific runner. Arbitrary commands → return byte-for-byte. Test `Agent: all` with three separate files and exact equality for the custom command.

internal/agent/grok.go
```diff
@@ -0,0 +1,597 @@
...
+	args = append(args, "--no-subagents")
+	args = append(args, "--disable-web-search")
+	return args
+}
+
+// appendGrokClassifySafetyArgs adds layered fail-closed classification flags.
+// Intentionally does not pass --tools "" (Grok treats empty as no override).
+func appendGrokClassifySafetyArgs(args []string) []string {
+	args = append(args, "--sandbox", grokSandboxReadOnly)
+	args = append(args, "--disallowed-tools", grokClassifyDisallowedTools)
```
EnzoTironi, line 303: **[suggestion] P1 — classify isolation is denylist-only; not structural**

Classify passes only `--disallowed-tools` (full default tool ID snapshot). Docs claim layered fail-closed behavior, but the boundary is a manual photograph of the registry (`SOURCE_REV at validation` is still a placeholder). New tools, plugin tools, or ToolKind aliases can escape.

Grok docs: non-empty `--tools` allowlist restricts the toolset but **retains MCP meta-tools**; when both flags are present, **denylist wins**, then allowlist filters the rest.

**Stronger shape:**

1. Non-empty positive allowlist seed (never empty `--tools`, which fail-opens)
2. Denylist = seed + MCP meta + known defaults (so seed and meta cannot remain)
3. Keep sandbox / no-subagents / disable-web-search / max-turns / no-memory / no-plan
4. Replace `SOURCE_REV` with the actually validated commit/release
5. Exact argv test (not merely "has a denylist")
6. Document residual drift without claiming Claude-equivalent deny-all

Remove production dependence on documenting empty-tools alone as the story.

internal/agent/grok.go
```diff
@@ -0,0 +1,597 @@
...
+	grokOutputFormatStreamingJSON = "streaming-json"
+	grokSandboxReadOnly           = "read-only"
+)
+
+// grokDefaultToolNames lists model-facing tool IDs from the default Grok Build
+// toolset (xai-org/grok-build registry + MCP meta). Used as a single source of
+// truth for fail-closed classification and residual mutating-tool denial.
+//
+// Source of truth: crates/codegen/xai-grok-tools ToolId::new registrations and
+// headless CLI flags --tools / --disallowed-tools (SOURCE_REV at validation).
```
EnzoTironi: **[suggestion]** `SOURCE_REV at validation` is still a placeholder. Pin the Grok release/commit whose tool registry was snapshotted, or drop the claim until it is real.

internal/agent/grok_test.go
```diff
@@ -0,0 +1,639 @@
...
+	  "additionalProperties": false,
+	  "required": ["design_review", "reason"],
+	  "properties": {
+	    "design_review": {"type": "boolean"},
+	    "reason": {"type": "string"}
+	  }
+	}`)
+
+	a := NewGrokAgent(grokPath)
+	ctx := context.Background()
```
EnzoTironi: **[suggestion] P1 — smoke test creates false confidence**

- Uses `context.Background()` despite the timeout comment
- Any `classifyErr` is treated as acceptable fail-closed (auth failure ≡ tools blocked)
- Only detects shell that creates a marker file; not `read_file` / `search_tool` / `use_tool`
- Does not parse `streaming-json` for `tool_call` events

**Fix:** real deadline; unexpected errors fail the opt-in smoke (caller assumed auth/network); capture stream and reject any `tool_call`; keep marker as a secondary signal only.

internal/agent/acp_agent.go
```diff
@@ -376,3 +388,32 @@ func (a *ACPAgent) runPrompt(
...
+// authenticateNonInteractive is intentionally a no-op in the shared ACP path.
+// Provider-specific method IDs (e.g. former Grok cached_token / xai.api_key)
+// are not hardcoded here; first-class Grok uses the native headless agent.
+func authenticateNonInteractive(
```
EnzoTironi: **[suggestion] P1 — dead ACP auth scaffolding**

`authenticateNonInteractive` always returns nil. `tryNonInteractiveAuth` is kept "for unit tests and future wiring". `initResp` is captured mainly to pass auth methods into a no-op. This PR's direction is first-class CommandAgent Grok and stripping Grok knowledge from shared ACP.

**Fix:** delete the no-op flow, helpers, `acp_noninteractive_auth_test.go`, unnecessary `initResp` capture, and metadata with no in-PR consumer — unless a real consumer lands in this contribution.

internal/agent/acp_noninteractive_auth_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// Shared ACP path does not hardcode provider auth method IDs.
+	require.NoError(t, authenticateNonInteractive(t.Context(), nil, nil))
+}
+
+func TestTryNonInteractiveAuth(t *testing.T) {
+	t.Parallel()
+
+	t.Run("noop when no candidates", func(t *testing.T) {
+		err := tryNonInteractiveAuth(nil, func(string) error {
+			t.Fatal("auth should not be called")
```
EnzoTironi: **[suggestion]** Uses `t.Fatal` inside the auth callback. AGENTS.md forbids `t.Fatal*` when a direct testify assertion covers the case. Prefer deleting this file with the no-op helpers; if kept, use `require`/`assert` only.

cmd/roborev/skills.go
```diff
@@ -195,6 +197,8 @@ This command is idempotent - running it multiple times is safe.`,
 						fmt.Println("  Codex: $roborev-review, $roborev-review-branch, $roborev-design-review, $roborev-design-review-branch, $roborev-fix, $roborev-respond")
 					case skills.AgentDroid:
 						fmt.Println("  Factory Droid: /roborev-review, /roborev-review-branch, /roborev-design-review, /roborev-design-review-branch, /roborev-lookahead-review, /roborev-lookahead-review-branch, /roborev-fix, /roborev-respond")
+					case skills.AgentGrok:
+						fmt.Println("  Grok Build: /roborev-fix, /roborev-refine, /roborev-respond")
```
EnzoTironi: **[suggestion] P1 — post-install "Try:" lists only 3 of 9 Grok skills**

Install ships the full Grok skill set (review / design / lookahead / fix / refine / respond, etc.). Output lists only `/roborev-fix`, `/roborev-refine`, `/roborev-respond`. Align with Claude/Codex/Droid completeness (or document intentional subset).

docs/advanced/acp.md
```diff
@@ -37,6 +37,10 @@ Set `command` to the wrapper binary in your ACP config:
 # ~/.roborev/config.toml
 [acp.codex-acp]
 command = "codex-acp"
+
+[acp.grok-custom]
```
EnzoTironi: **[suggestion] P2 — `grok-custom` ACP example fights the PR direction**

This contribution makes Grok a native CommandAgent and removes Grok-specific policy from shared ACP. Re-adding a TOML example of `command = "grok"` + ACP stdio reopens the surface the cleanup is closing. Remove the example.

internal/agent/identity.go
```diff
@@ -0,0 +1,132 @@
...
+	return false
+}
+
+func runIdentityProbe(ctx context.Context, command string, flag string) ([]byte, error) {
+	cmd := exec.CommandContext(ctx, command, flag)
+	// Isolate from the parent tty and env noise; no stdin.
+	cmd.Stdin = nil
+	// On Windows, copies of grok.exe as agent.exe still answer --version.
+	if runtime.GOOS == "windows" {
+		cmd.SysProcAttr = nil
```
EnzoTironi: **[nit]** `cmd.SysProcAttr = nil` is a no-op (`nil` is already the zero value). Removes neither a Windows quirk nor documents real behavior. Drop the branch and the unused `runtime` import if that was the only use.

internal/agenthook/install.go
```diff
@@ -105,6 +117,29 @@ func RunInstall(opts InstallOptions, stdout io.Writer) error {
 		}
 		printInstallResult(stdout, "Claude", path, changed, opts.DryRun)
 	}
+	if agent == "all" || agent == "grok" {
+		// "all" uses the Codex/Claude plain runner command; rewrite its suffix
+		// to the Grok-specific runner so install never clobbers shared entries.
+		grokCommand := command
+		if agent == "all" {
+			grokCommand = rewriteHookRunner(command, grokAgentHookRunner)
+		}
```
EnzoTironi, line 126: **[bug] P0 — call site that rewrites for Grok under `agent == "all"`**

`rewriteHookRunner(command, grokAgentHookRunner)` is correct for roborev-generated runners, but unsafe when `command` is a user `--command` override without `agent-hook run` (appends the runner). See comment on `rewriteHookRunner` fallback.

internal/streamfmt/streamfmt.go
EnzoTironi: Fixed in fe30187: Message is json.RawMessage; Grok error strings decode to formatted error: <msg> without dumping raw JSON. Negative assertions cover the official shape.

EnzoTironi:
## Push + re-review note (`296451f`)

Pushed five fix commits on top of `c5ec550`:

1. `fe30187` — streamfmt native Grok event shapes  
2. `6c811c0` — Cursor identity on every command path  
3. `c41bdf6` — structural classify isolation  
4. `4a0c701` — preserve custom hook commands  
5. `296451f` — ACP cleanup + skills/quickstart UX  

Local gates green; live `ROBOREV_RUN_GROK_SMOKE=1` smoke passed.

**CI:** CI + Windows Migration on this head are `action_required` (fork workflow approval). Needs a **kenn-io maintainer** to approve:

- https://github.com/kenn-io/roborev/actions/runs/30777889740  
- https://github.com/kenn-io/roborev/actions/runs/30777889734  

Pending re-review posted on `296451f` (supersedes the incomplete `c5ec550` review). PR remains **draft** until CI is green.

cmd/roborev/review.go
EnzoTironi: Keeping Droid in the `review --help` agent enumeration is intentional.

Droid was already a first-class registered agent; this PR only extended the help list for Grok. Omitting Droid while touching that line would leave help knowingly incomplete relative to the registry. Same correction pattern as documenting any other already-registered agent when the enum is edited.

internal/agent/agent_test_assert_helpers_test.go
EnzoTironi: Keeping Pi in `expectedAgents` is intentional.

That list is the complete registered-agent source of truth for tests. Pi was already registered; adding Grok without Pi would break the invariant that the helper matches the real registry. This is restoring completeness, not expanding scope beyond agents that already ship.

EnzoTironi:
Pushed `0289ad6` + `627571c` on head `627571c`: Cursor identity probes are side-effect-free (flags only, tri-state, fail-closed). Superseding pending review replaces the "CI only" note — that was premature. Still draft; Linux/Windows CI need maintainer workflow approval.

EnzoTironi (commented):
## Re-review of head `8972777` (probe bounds + Windows fixtures + auditable pin)

Previous pending review on `627571c` is **superseded** — do not submit it.

### Blockers fixed on this head

| Issue | Fix |
|-------|-----|
| Probe not truly bounded (`CombinedOutput` + `WaitDelay==0`) | `configureCapabilityProbe`, `WaitDelay=250ms`, explicit `Cancel`, capped concurrent stdout/stderr writer (`identityProbeMaxOutput`), `ErrWaitDelay`/oversize → `identityUnknown` |
| Orphan pipes could hang availability | Regression test: parent + background child; returns within `timeout+WaitDelay+margin` |
| Windows `.bat` `if … & exit` syntax | Parenthesized `if (...)` blocks; dedicated `//go:build windows` fixture test for `--version`/`-v`/positionals |
| Fail-open cache of unknown | Unchanged: only conclusive Grok/NotGrok cached |
| Prompt positionals | Unchanged: only `--version`/`-v` |
| Registry pin not auditable | Split: `grokToolsCLIVersion`, `grokToolsBinaryIdentity`, `grokToolsRepoRev` (public `xai-org/grok-build`), `grokToolsMonorepoRev` (`SOURCE_REV`) |
| "Genuine Cursor" naming | Renamed to `commandIsUsableCursorCandidate` (not-Grok + conclusive version only) |
| Docs | **Prompt-safe** (not "side-effect-free"); `cursor_cmd` must forward version flags |

Commits: `bc3abd9` (bound I/O + Windows fixtures + rename), `8972777` (auditable pin + docs).

### Still correct from prior stack

streamfmt native Grok shapes; structural classify seed+denylist; custom hooks preserved; ACP dead auth removed; skills/quickstart; positional-free identity; fail-closed unknown.

### Local verification

- `go fmt` / `go vet` / `go build ./...`
- `go test ./internal/agent ./internal/streamfmt ./internal/agenthook` (+ `-race` on those packages)
- `make lint-ci`, `make markdown-ci`
- Pre-commit hooks green on both commits

### Still open

**Linux + Windows CI** remain maintainer-gated (`action_required` on fork workflows). Windows Migration is especially important for the `.bat` fixture fix. Local green does not replace that.

### Recommendation

- Keep PR **draft** until Linux + Windows CI run green on this head.
- Do **not** submit older pending reviews that claim “only CI remains” without the probe-bound and Windows fixture fixes.

wesm:
Looking at this

wesm (approved):

wesm:
fixing windows CI

wesm:
Valid underlying concern, but this is not Grok-specific. Addressing only Grok would create inconsistent reviewer capabilities without fixing the shared credential boundary. Cross-agent isolation and credential separation are tracked in #1022

PR #1019, opened 2026-08-03
feat(agent-hook): add workspace-scoped snoozing

mariusvniekerk:
Agent Hook reminders are useful, but long implementation stretches need a quieter mode. Existing queue pausing is too broad because it stops the asynchronous reviews users still want running.

This adds `roborev snooze` / `roborev snooze on --duration <duration>` and `roborev snooze off`, with an eight-hour default. Snooze state is keyed to the current repository, linked worktree, and branch. It suppresses only Agent Hook reminders while reviews continue enqueueing, processing, and accumulating findings; hook baselines advance during the quiet period to avoid catch-up noise afterward.

The state remains local SQLite-only because worktree paths and snooze intent are machine-specific. Bundled Codex, Claude, and Droid skills expose both on and off operations, with machine-readable metadata that keeps snoozing human-triggered only. The README plus Agent Hook and agent-skills guides document the behavior.


cpcloud (approved):
Lotta code :)

PR #269, opened 2026-02-16
Fix Windows 32KB command-line limit for agent prompts

wesm:
## Summary

- Switch copilot, cursor, droid, and opencode agents from CLI arguments to stdin piping for prompt delivery, avoiding the Windows `CreateProcess` 32,766-character limit
- Switch `gh pr comment` from `--body` to `--body-file -` with stdin piping for the same reason
- Add `cmd.WaitDelay = 5s` to all pipe-based agents (claude, codex, cursor, gemini) so context cancellation reliably terminates pipe reads when child processes outlive the parent
- Add `--large-prompt` flag to `check-agents` for validating the fix with a 33KB+ prompt

## Test plan

- [x] `go test ./...` passes
- [x] `roborev check-agents` — all agents pass
- [x] `roborev check-agents --large-prompt` — agents that support large prompts pass

Closes #266

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1023, opened 2026-08-05
fix(agent): pass OpenCode prompts as arguments

shunkakinoki:
## Summary

OpenCode's `run` command expects the review message as its positional argument. Passing the prompt only through stdin leaves the session without a message and exits unsuccessfully on current OpenCode releases.

This change passes the prompt as the positional `run` message and keeps stdin unused for the invocation.

## Tests

- `go test ./internal/agent`
- `go test -tags=integration ./internal/agent -run TestOpenCodeReviewPassesPromptAsArgument -count=1 -v`
- Verified the JSON stream path with a real OpenCode invocation using an explicit model.

### DIFF
--- a/internal/agent/opencode.go
+++ b/internal/agent/opencode.go
@@ -99,14 +99,13 @@ func (a *OpenCodeAgent) Review(
 	repoPath, commitSHA, prompt string,
 	output io.Writer,
 ) (string, error) {
-	args := a.buildArgs()
+	args := append(a.buildArgs(), prompt)
 
 	runResult, runErr := runStreamingCLI(ctx, streamingCLISpec{
 		Name:    "opencode",
 		Command: a.Command,
 		Args:    args,
 		Dir:     repoPath,
-		Stdin:   strings.NewReader(prompt),
 		Output:  output,
 		// opencode prints sqlite-migration progress to stderr
 		// on every invocation, drowning the live log with
--- a/internal/agent/opencode_test.go
+++ b/internal/agent/opencode_test.go
@@ -55,16 +55,15 @@ func TestOpenCodeReviewModelFlag(t *testing.T) {
 	}
 }
 
-func TestOpenCodeReviewPipesPromptViaStdin(t *testing.T) {
+func TestOpenCodeReviewPassesPromptAsArgument(t *testing.T) {
 	t.Parallel()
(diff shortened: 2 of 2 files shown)


mariusvniekerk:
I tried to reproduce the premise of this PR and could not, on either opencode 1.17.7 or 1.18.13 (the latest release at time of writing), macOS arm64, using a fresh isolated data dir (`XDG_DATA_HOME` pointed at an empty directory):

```
$ printf 'Reply with the single word OK and nothing else.' | opencode run --format json
...
{"type":"text",...,"part":{...,"type":"text","text":"OK",...}}
{"type":"step_finish",...,"reason":"stop",...}
$ echo $?
0
```

The prompt is received via stdin and the run exits successfully, on both versions. So "passing the prompt only through stdin leaves the session without a message and exits unsuccessfully on current OpenCode releases" does not hold — `opencode run` reads stdin as the message when no positional argument is given.

The stdin transport is also deliberate, not incidental: #269 (2bd3649) moved agent prompts off argv because Windows caps the command line at 32KB. Passing the prompt as a positional argument reintroduces that, and on Linux hits the ~128KiB per-argument limit (`MAX_ARG_STRLEN`) — while roborev's default max prompt size is 250KB, so ordinary review prompts would fail at exec before opencode even starts. It also exposes the full review prompt (diffs included) in process listings, as the CI review noted.

If stdin runs are failing for you locally, could you share your opencode version and the stderr output? One failure mode I did reproduce is a corrupted local opencode state DB (schema mismatch after a version switch), which fails with a sqlite error at `SessionPrompt.createUserMessage` — identically for both the stdin and positional forms, so this change wouldn't help in that situation either.

shunkakinoki:
You're right. I reproduced stdin successfully on the same machine with OpenCode 1.18.10 on macOS arm64:

```sh
printf 'Reply with the single word OK and nothing else.' | opencode run --format json
```

That returns a `text` event containing `OK` and exits successfully. The original failures used a different provider/model, and the OpenCode log shows `AI_APICallError: Weekly usage limit reached`. After switching models, stdin works as intended. I misattributed the provider failure to the stdin transport.

The argv change would reintroduce the prompt-exposure and command-size problems addressed by #269, so I'm closing this. Thanks for the careful reproduction.

PR #1034, opened 2026-08-11
Add kata-style split-screen review view to the TUI

gramster:
- Adds a split-screen layout to `roborev tui` on terminals >= 140x36: queue list in a left pane, review details in a right pane, modeled on kata's split view (shared full-width chrome, bordered panes,
  focused-pane accent border)
  - `L` toggles split/stacked with a sticky preference that survives resizes; below the breakpoint the TUI degrades to the existing full-screen layout, which is unchanged
  - The detail pane auto-follows the queue cursor with a 75ms debounced fetch and generation guards, so holding j/k coalesces into one fetch and stale responses are dropped
  - The pane is always useful: done jobs show the review, running jobs show a status card with a live log tail that auto-swaps to the review on completion, failed jobs show the error, queued jobs show a metadata
  card; fetch failures render in-pane with a retry hint
  - Focus model: `tab` into the detail pane, `esc` back to the list; all existing review keys work in the pane; transient views (prompt, comment, help, log, tasks) stay full-screen and round-trip back to the
  split; Enter in the split queue is a no-op since the pane already follows
  - Mouse: click to focus/select by pane, wheel scrolls the pane under the cursor
  - Queue columns auto-narrow in the list pane (dropping from the end of the configured column order; core columns always survive), reusing the existing column-config system
  - Control socket `get-state` now reports `layout` and `focus`; docs and help view updated
  - All new behavior is gated behind the split layout; stacked-mode rendering is untouched and the full test suite passes

### DIFF
--- a/cmd/roborev/tui/action_test.go
+++ b/cmd/roborev/tui/action_test.go
@@ -492,6 +492,10 @@ func TestTUIClosedRollbackRestoresSelectionAfterLeavingQueue(t *testing.T) {
 	assertSelection(t, m, 2, 3)
 
 	m.currentView = viewReview
+	// viewReview requires a loaded review (normalizeSplitState repairs a
+	// dangling viewReview/nil-currentReview pair back to viewQueue in
+	// every layout); this test only cares about selection rollback.
+	m.currentReview = &storage.Review{JobID: 3, Job: &m.jobs[2]}
 
 	m, _ = updateModel(t, m, closedResultMsg{
 		jobID:            2,
--- a/cmd/roborev/tui/actions.go
+++ b/cmd/roborev/tui/actions.go
@@ -282,7 +282,10 @@ func (m model) submitComment(jobID int64, text string) tea.Cmd {
 			return commentResultMsg{jobID: jobID, err: fmt.Errorf("submit comment: %w", err)}
 		}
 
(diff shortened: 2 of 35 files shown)


gramster:
I am parking this for a week. Using way too much of my weekly token budget. I think it will be good for PR #1038 to land too; I see that that attempts to reduce "severity-inflated rare races", much of which has been consuming the effort here (and I think to the detriment of the code, as increasingly elaborate mitigations for these are being added).

FWIW @mariusvniekerk  in the code review tool we built on my team (which I told Wes about), we had several different personas, including a nit-picking "skeptic" and an "advocate" who would counter the skeptic, and all of them submitted their reviews to an "adjudicator" - there was some back and forth to try to drive consensus and the end result was something like an Supreme Court opinion where there was a majority view and (in some cases) a couple of places where there was still dissent. Obviously this burned up a bunch more tokens than a single review agent would - we were fortunate to have an unlimited budget - but it did largely eliminate things like severity inflation.

wesm:
I can pick this up and finish it if that's cool, I have lots of tokens

wesm:
I'll need to do some QA on this before merging

mariusvniekerk:
image

mariusvniekerk:
works nicely

PR #1038, opened 2026-08-11
Calibrate the security review prompt: settled non-findings, precision posture, comparative pass

mariusvniekerk:
## Summary

- Add settled non-findings to the security review prompt's do-not-report list: denial of service and resource exhaustion, log spoofing (logging remains a finding when it exposes secrets or PII), regex injection and regex-based DoS, memory-safety claims in memory-safe languages, findings confined to test-only files, and untrusted content reaching an AI prompt unless the model's output can trigger a privileged action the content's author could not already perform.
- State the precision posture explicitly — better to miss a theoretical issue than flood the report; each finding should be one a security engineer would confidently raise.
- Add a comparative pass: judge changed code against the codebase's established secure pattern for the same class of operation. A new deviation from that pattern is a strong finding; consistency with it usually is not.
- Motivation: these are the noise classes repos keep suppressing one-by-one in per-repo `review_guidelines` — kit's history records the CI panel repeatedly flagging the same safe.directory pattern until a guideline shut it down, and roborev's own guidelines devote a page to excluding local-only trust-model findings. Adjudications that generalize belong in the default prompt; genuinely repo-specific trust models stay in `.roborev.toml`.
- Scope note: the dedicated `security` review type has almost no usage history (two jobs ever recorded), so this calibrates against security-flavored noise observed in general/design reviews. The existing exploitability-burden-of-proof structure, attacker/boundary verification questions, and severity definitions are unchanged.

### DIFF
--- a/internal/prompt/templates/default_security.md.gotmpl
+++ b/internal/prompt/templates/default_security.md.gotmpl
@@ -2,7 +2,9 @@ You are a security code reviewer with an exploitability burden of proof. Review 
 
 Report an issue only when the changed code affects a real trust boundary, security decision, secret-bearing path, privileged operation, or externally/user-controlled input path. The finding does not need a turnkey exploit, but it must identify a realistic attacker capability, the weakened boundary or control, and the concrete asset or privilege at risk.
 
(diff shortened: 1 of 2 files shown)


wesm:
Seems fine to me. Need to get full CI

PR #1043, opened 2026-08-13
fix(codex): map GPT-5.6 maximum to max

salmonumbrella:
Stacked on #1042, which introduces distinct exact `xhigh` and `max` reasoning tiers.

RoboRev's legacy `maximum` preset should request the highest compatible effort. GPT-5.6 `sol`, `terra`, and `luna` distinguish `xhigh` from `max`, so this maps `maximum` to literal `max` for those explicit model aliases while retaining `xhigh` for older, default, and unknown Codex models. Exact `xhigh` remains a literal `xhigh` request on every model.

The model allowlist is deliberately conservative: new or custom model IDs keep the backward-compatible ceiling until their `max` support is known.


cpcloud:
Thanks, merging this into mine.

PR #1044, opened 2026-08-13
Batch automatic post-commit reviews

salmonumbrella:
roborev currently queues a review after every commit. That gets noisy fast on repos with lots of small commits.

This adds a repo-local `post_commit_batch_size`, so the post-commit hook waits for N commits and enqueues one review over the accumulated range. Below the threshold the hook returns without starting or contacting the daemon, so sub-threshold commits stay cheap. Commit mode reviews `checkpoint..HEAD`; branch mode keeps the full branch scope, just at the configured frequency. Disabling batching drains the current range instead of dropping it.

Checkpoints are kept per branch in a JSON file under the shared git common dir, so linked worktrees agree on the same pending range. Hook runs are serialized with a file lock, and a checkpoint advances only after a successful enqueue, so a failed enqueue retries the same range on the next commit. A new pre-push hook flushes pending work for pushed `refs/heads` branches before they leave the machine.

There is deliberately no history-rewrite tracking: after a rebase or amend the checkpoint recovers from the merge base and re-reviews from there, and any other error fails open to an immediate single-commit review. The worst case is a redundant review, never a silently skipped commit.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,53 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '37 4 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      packages: read
+      security-events: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - language: actions
+            build-mode: none
+          - language: go
+            build-mode: manual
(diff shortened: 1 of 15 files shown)


mariusvniekerk:
One thing we can potentially also do is for the markers that the post-commit receives to basically go into a cooldown / commit count period.  Since the pure n commit form can easily prevent the last commit of a PR from being reviewed.

wesm:
looking at this

wesm:
declined. merging

wesm:
some kind of a force push failure, I'm opening a new PR

wesm:
Superseded by #1088, which carries the same squashed commit (original authorship preserved). This PR was accidentally closed by a faulty force-push that emptied the head branch.

PR #1056, opened 2026-08-15
Constrain Agent Hook autofix reminders to exact reviews

mariusvniekerk:
Agent Hook reminders could repeatedly rediscover the same failed reviews and turn background feedback into unrelated repair work. The default fallback could also start a separate fix agent that had no reliable view of the user's active task.

Reminders now name the exact review jobs they surfaced and invoke only the bundled `roborev-fix` skill. The skill treats findings as claims to prove against current code: it fixes valid in-scope problems, closes invalid reviews with evidence and no code change, and leaves valid out-of-scope work open for the user. Customized instructions remain complete overrides.

Delivered job IDs are remembered per agent session and repository lineage so later hook events do not reopen the same loop; newly created jobs remain eligible. Observation stays best-effort because post-tool hooks run frequently. Hook installation also refreshes bundled skills for Claude Code, Codex, Factory Droid, and Grok Build. Profiles without a bundled variant deliberately receive no generic CLI fallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,7 @@ your agentic loop while context is fresh.
 
 ```bash
 roborev init                  # layer 1: per-commit reviews
-roborev skills install
-roborev agent-hook install    # layer 2: auto-detect and wire installed agents
+roborev agent-hook install    # layer 2: wire agents and bundled skills
 roborev agent-hook install --agent all  # or wire every supported profile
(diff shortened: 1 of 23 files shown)


mariusvniekerk:
The CLI Fallback is rejected since its a bad idea.

PR #1057, opened 2026-08-15
Keep branch reviews in the invoking linked worktree

wesm:
## Summary

- Bind `roborev review --branch` to the linked worktree that contains the invocation instead of trusting a shared `core.worktree` top-level path.
- Verify the local `.git` marker against the worktree-specific Git directory before selecting the repository, branch, and commit range.
- Prevent workspace tooling or stale shared Git configuration from silently enqueueing a review for a sibling checkout.

### DIFF
--- a/cmd/roborev/review.go
+++ b/cmd/roborev/review.go
@@ -106,7 +106,7 @@ Examples:
 			}
 
 			// Get repo root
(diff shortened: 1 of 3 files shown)


wesm:
I tested this and this confirmed branch reviews that were queued incorrectly from a sibling worktree

PR #1067, opened 2026-08-17
Fix streamed Grok text wrapping in TUI logs

cpcloud:
Grok Build emits assistant text as adjacent streaming chunks. The TUI treated each chunk as a complete Markdown document, which inserted a line break after every chunk and could reduce a wide log pane to one or a few words per row.

This change keeps adjacent Grok text together until a meaningful stream boundary and carries that buffered response across incremental polling in both the full-screen log and split detail pane. Running polls keep the formatter open; the terminal poll finalizes it even when the daemon has no new bytes to return.

The assembled response still goes through the existing width-aware Markdown renderer, so genuinely narrow panes continue to wrap normally. Non-Grok events and plain stderr remain response boundaries.

### DIFF
--- a/cmd/roborev/tui/fetch.go
+++ b/cmd/roborev/tui/fetch.go
@@ -1050,8 +1050,10 @@ func (m model) fetchJobLog(jobID int64) tea.Cmd {
 			isIncremental = false
 		}
 
-		// No new data — return early with current state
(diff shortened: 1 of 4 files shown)


cpcloud:
Closing this out in favor of a stream format refactor, shortly inbound.

PR #1070, opened 2026-08-18
Coordinate self-updates with active reviews

wesm:
## Why

`roborev update` could replace the binary and then appear stuck while graceful
daemon shutdown waited for long-running reviews. The updater did not surface
that wait before installation or let the user choose how active work should be
handled.

## What changes

- Add `--running=wait|interrupt|abort` and a single interactive policy prompt
  when reviews are active. Non-interactive updates keep the safe `wait` default.
- Prepare updates through a renewable, daemon-owned drain lease. New reviews can
  still be enqueued, while the existing claim gate prevents queued work from
  starting before the replacement daemon is ready.
- Requeue update-interrupted attempts without consuming a retry or firing normal
  cancellation, panel, hook, cooldown, or failover side effects. Attempt-scoped
  storage guards preserve a concurrent user cancellation.
- Verify that the replacement daemon is responsive and reports the installed
  version before declaring success. Older daemons retain a policy-specific
  compatibility path.
- Present stable phase-oriented update output and expose active or recovering
  update drains through `roborev status`.

```mermaid
flowchart LR
    U[Updater] -->|Prepare lease| D[Daemon]
    D --> G[Block new claims]
    D --> P{Running policy}
    P -->|wait| F[Finish active reviews]
    P -->|interrupt| Q[Requeue active attempts]
    F --> I[Install and restart]
    Q --> I
    I --> V[Verify daemon version]
```

## Design notes

The drain begins before download because the existing installer deliberately
couples download, checksum verification, archive validation, and atomic
installation. Lease renewal and expiry keep that longer drain window
recoverable without adding an updater-side operation file or data-directory
lock.


wesm:
Addressed both reliability findings in `851ad1bd`:

- An update-target marker is now authoritative before context cancellation is
  observed. A targeted attempt cannot enter normal retry or failover handling
  during the marker-to-cancel interval, and the attempt-scoped database guard
  still lets a concurrent user cancellation win.
- Old-daemon exit now waits until the updater context is canceled. The 10-second
  timeout applies separately to replacement-daemon readiness, so legitimate
  hook and cleanup time no longer causes a premature restart failure.

The PR is larger than the hand-written feature alone. Of roughly 4,972 added
lines before these fixes, 1,984 are tests and 1,242 are generated OpenAPI or
client artifacts. The generated changes are required by `make api-check` for
the three update-drain endpoints and the checked-in Go and TypeScript clients.
The remaining roughly 1,688 lines are the CLI, daemon, worker, storage, and
documentation implementation. I kept the generated artifacts and concurrency
coverage because removing either would leave the public API stale or weaken the
cutover guarantees this change introduces.

wesm:
Addressed in `bc2ad3f5` with one worker-pool transition barrier:

- Interrupt preparation takes the exclusive side across the persisted claim
  gate, the running-job snapshot, and target registration/cancellation.
- Retry, failover, failure, normal completion, classifier, and synthesis paths
  hold the shared side from the marker check through their attempt-scoped
  database transition and related side effects.

The resulting order is explicit: either a worker transition finishes first and
the subsequent update snapshot excludes that attempt, or update preparation
registers the target first and the worker requeues it without consuming a
retry. A regression blocks an agent-error transition at the former race window
and verifies that interrupt preparation cannot publish its target until that
transition finishes; it also passes under the race detector.

PR #1088, opened 2026-08-22
Batch automatic post-commit reviews

wesm:
roborev currently queues a review after every commit. That gets noisy fast on repos with lots of small commits.

This adds a repo-local `post_commit_batch_size`, so the post-commit hook waits for N commits and enqueues one review over the accumulated range. Below the threshold the hook returns without starting or contacting the daemon, so sub-threshold commits stay cheap. Commit mode reviews `checkpoint..HEAD`; branch mode keeps the full branch scope, just at the configured frequency. Disabling batching drains the current range instead of dropping it.

Checkpoints are kept per branch in a JSON file under the shared git common dir, so linked worktrees agree on the same pending range. Each entry records its pending range — the last reviewed boundary and the branch head last observed — so a range is tied to a history, not just a branch name. Hook runs are serialized with a file lock, and a checkpoint advances only after a successful enqueue, so a failed enqueue retries the same range on the next commit. A new pre-push hook flushes pending work for pushed `refs/heads` branches before they leave the machine; a pushed branch with no checked-out worktree is flushed from the pushing worktree, and explicit branch flushes run even during a rebase.

There is deliberately no reflog or rename tracking: after a rebase or amend the checkpoint recovers from the merge base and re-reviews from there, and any other error fails open to an immediate single-commit review. A rename leaves its entry orphaned and the branch whose first-parent chain holds the recorded range adopts it; a branch name reused after a rename hands the old range to the branch that now holds it. The worst case is a redundant review, never a silently skipped commit.


wesm:
Replaces #1044 (accidentally closed by a faulty force-push to its fork branch; fork push access was lost when it closed). Same squashed commit, review history lives in #1044.

Issue #1011, opened 2026-07-30
roborev can report a clean verdict without having reviewed the requested content (4 issues confirmed on v0.61.2)


pmcfadin:
## Context

We drive `roborev` from automated agents as a **merge gate** — a review must be confirmed clean before a PR
is allowed to land. That use case makes us unusually sensitive to one failure mode: **roborev returning a
plausible-looking clean verdict when it did not actually review the requested content.** A vacuous
`No issues found.` is textually identical to a genuine clean pass, so an automated consumer cannot tell them
apart without out-of-band evidence.

This is not hypothetical for us. On one of our changes, two runs reported "No issues found"; re-run correctly
against the right commit, the **same diff produced two genuine blocking findings**. Accepting the first
verdict would have merged unreviewed code.

We ended up building a defensive wrapper that verifies roborev's output against independently-computed facts.
In doing so we found the following. Everything in §1 I reproduced directly on **v0.61.2**; §2 is honestly
flagged as older evidence I have not re-verified; §3 is a doc/UX note, not a bug.

Happy to split this into separate issues if you'd prefer — each section is independently actionable. Also
happy to test patches; we exercise this heavily.

**Environment:** roborev **v0.61.2** (static release), Linux x86_64, `--agent codex --model gpt-5.6-sol`,
daemon healthy, 4 workers. Clean database (`roborev repo list` was empty at session start), so the job ids
below are a fresh 1 and 2.

---

### 1a. `--wait` returns before the job record is durable

`roborev review <sha> --wait` returns, but the job record is **not yet readable**. Immediately after the
command exited, `roborev list --json` for job 2 had **no `git_ref`, no `status`, no `model`, and no
`token_usage`**. Queried again moments later, all four were populated:

```
id=2 status='done' git_ref='155e12c8e8d1af9f75e47ca854e87f489b818fe6' model='gpt-5.6-sol' token_usage=<present>
```

**Repro:** run `roborev review <sha> --repo <abs> --agent codex --model gpt-5.6-sol --wait`, then immediately
`roborev list --json` (or `roborev show <job> --json`) in the same shell. Re-query a few seconds later.

**Impact:** a consumer that reads the record right after `--wait` — the natural thing to do — silently gets a
degraded view. In our case four independent verification signals fell back to weaker sources simultaneously,
with no error and no indication anything was missing. We now poll with a bounded retry, but ideally `--wait`
would not return until the record it produced is readable (or would expose an explicit "record durable"
signal).

[...]

PR #1137, opened 2026-09-04
feat(review): record and display review file coverage

rodboev:
The review daemon now records file coverage for committed and range reviews. It stores reviewed and excluded file counts on each review, and `roborev show`, the review API, and terminal review details display those values after completion.

Previously, a completed review stored its verdict without recording how much of the change set reached the reviewer. That left later readers unable to distinguish a review that received the full change set from one whose path filters removed files. Coverage now comes from the same Git path enumeration used for the review input, so the stored result remains tied to the historical review.

## Details

- The worker calls `DiffPathsCtx` with and without the review's exclusion pathspecs. Reviewed files are unique retained paths; excluded files are paths present only in the full change set.
- Review completion writes nullable `reviewed_file_count` and `excluded_file_count` columns. A measured zero stays distinct from unknown coverage.
- SQLite and PostgreSQL sync use `COALESCE` when updating existing reviews so incoming records with missing counts preserve stored values.
- Path enumeration failures leave coverage unknown and allow the review to continue.

Prompt construction remains unchanged. Dirty reviews, prebuilt inputs, prompt-native jobs, fix jobs, and panel synthesis remain without coverage. Browser projections omit the coverage fields, while synchronized review records preserve them. A rename counts as two paths.

Refs #1011


mariusvniekerk (changes requested):
Requesting changes: this solves the wrong problem, and the design is fundamentally flawed because it puts human-facing review metadata into the agent's prompt.

Helping people see what a review covered is useful. This implementation instead changes the input sent to the review agent to make that prompt easier for humans to inspect. The existing prompt already identifies the review subject; adding included/excluded file counts provides no demonstrated benefit to the agent and consumes prompt budget that could otherwise hold review content. It adds noise and risks making reviews worse.

If we want to capture additional information about the changed files, store it as structured metadata on the review/job and display it in the CLI or TUI review details. Human-facing scope information should not require changing the actual prompt contents. Please redesign this around metadata and presentation rather than adding these annotations to the review prompt.

mariusvniekerk (approved):
The redesign addresses my original objection: file coverage is now stored as structured review metadata and displayed in human-facing review details, without adding it to the agent prompt. Approving on that design basis.

PR #1158, opened 2026-09-09
Require evidence for build and test failure findings

wesm:
Built-in reviewers are asked to reproduce compile and test-failure claims with the repository-pinned toolchain, include the command and failure output, and label unavailable checks as unverified.

They are also asked to trace complete fixture setup before reporting missing initialization. These are review instructions, not an enforced proof requirement or a claim of measured accuracy improvement.

### DIFF
--- a/docs/guides/reviewing-code.md
+++ b/docs/guides/reviewing-code.md
@@ -3,6 +3,19 @@ title: Reviewing Code
 description: Review branches, uncommitted changes, and commit ranges
 ---
 
+## Evidence for Build and Test Findings
+
+The built-in review prompts ask reviewers to run a focused check before claiming
+that code does not compile or a test fails. Reviewers should use the
+repository's pinned toolchain at the reviewed revision and include the command,
+version, and failure output. If execution is unavailable, they should label the
+claim unverified and explain the limitation.
+
+For missing setup findings, reviewers should trace the complete initialization
+path. A helper, export, constructor, or transaction callback may provide state
+that is not visible beside the failing operation. These instructions guide the
+reviewer; they are not a guarantee that every finding has been reproduced.
+
 ## Feature Branches
 
 Use `--branch` to review all commits since your branch diverged from main:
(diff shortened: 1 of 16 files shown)


mariusvniekerk (changes requested):
Requesting changes: the review instructions contradict each other about whether reviewers may run code.

The new instruction in [internal/prompt/prompt.go:71](https://github.com/kenn-io/roborev/blob/423a05e97f8a517e03271d860039cfe0e96e4143/internal/prompt/prompt.go#L71) says to "run the smallest relevant check" before claiming that code does not compile or a test fails.

But all three agent review templates say: "Do NOT build the project, run the test suite, or execute the code while reviewing."

- [Codex template, line 5](https://github.com/kenn-io/roborev/blob/423a05e97f8a517e03271d860039cfe0e96e4143/internal/prompt/templates/codex_review.md.gotmpl#L5)
- [Claude template, line 5](https://github.com/kenn-io/roborev/blob/423a05e97f8a517e03271d860039cfe0e96e4143/internal/prompt/templates/claude-code_review.md.gotmpl#L5)
- [Gemini template, line 5](https://github.com/kenn-io/roborev/blob/423a05e97f8a517e03271d860039cfe0e96e4143/internal/prompt/templates/gemini_review.md.gotmpl#L5)

Both instructions reach the reviewer in the same prompt. For example, the [assembled Codex prompt](https://github.com/kenn-io/roborev/blob/423a05e97f8a517e03271d860039cfe0e96e4143/internal/prompt/testdata/golden/single_review_codex.golden) contains both. The reviewer cannot follow both when a suspected failure needs a build or test to verify it.

I recommend allowing only small, focused build or test checks needed to verify a specific suspected failure, using the repository-pinned toolchain. Reviewers should not run the full test suite or broad project builds. If a focused check cannot be run, the failure claim should remain explicitly unverified.

Please make the shared instruction and all three agent templates consistent with that scope, and update the documentation and expected prompt output to match.

PR #1163, opened 2026-09-12
fix: prefer archived usage while supporting older AgentsView CLIs

shunkakinoki:
Prefer `agentsview session usage --no-sync` so supporting CLIs read archived usage, including subagents, without synchronizing source transcripts.

Keep existing installations working: when `--no-sync` is unknown, retry `session usage` without it; when `session usage` is unavailable, use the legacy `token-use` command. These fallback commands may synchronize sources. Cost-capable versions continue reporting costs, and older versions still provide token counts. Provider failures remain errors, and missing sessions or absent usage still return no data for existing retry handling.

The [companion AgentsView change](https://github.com/kenn-io/agentsview/pull/1739) enables the preferred archive-only path, but this PR can land independently. The configured HTTP endpoint path is unchanged.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1379,7 +1379,12 @@ default_model = "anthropic/claude-opus-4-8"
 ### Cost Usage Endpoint
 
 By default, roborev looks up token usage and cost estimates through the local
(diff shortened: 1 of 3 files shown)


wesm:
looking

PR #1188, opened 2026-09-16
Detect Windows Terminal background via OSC 11 for color-mode auto

### LINKED ISSUE #1176: TUI automathic theme issue
Hi I am reading https://www.roborev.io/docs/configuration/#color-mode and it seems this is an edge case I have. I use a Win 11 powershell with the application color sheme set to solarized light. The TUI however seems to default to dark and I have to run  with 'ROBOREV_COLOR_MODE=light roborev tui'

I guess this is an edge case in how color mode detects what to run. Not a big deal, but thought good to report it.

novica:
Use Windows Terminal's reported background for automatic colors in the TUI
and standalone commands such as `log`, `fix`, and `refine`. The existing
termenv detector reports black on Windows, so light themes receive the dark
palette.

The TUI initializes its styles without reading terminal input. Bubble Tea
requests the background after its input loop starts and applies replies to
adaptive colors, cached Markdown, and logs. Late replies use that same loop.

Standalone commands make one foreground OSC 11 attempt per process when both
streams are console handles and no input is queued. They peek until a complete
BEL- or ST-terminated color reply is available, then remove only its records.
The whole attempt uses [termenv v0.16.0's five-second OSC timeout](https://github.com/muesli/termenv/blob/v0.16.0/termenv_unix.go#L16-L19),
so partial replies cannot extend detection indefinitely. Failed detection keeps
the dark palette; explicit color modes and `NO_COLOR` skip the query.

Standalone detection remains best effort: a reply arriving after detection stops can
remain queued for the next reader. Forced `dark` or `light` mode avoids that
query. Native Windows Terminal behavior still needs confirmation.

Fixes #1176


mariusvniekerk:
@rodboev can you take a windows look here plz.

rodboev:
This freezes the TUI in Windows Terminal. Line 62 is the problem: `reply, err := bufio.NewReader(os.Stdin).ReadString('\a')`

This goroutine is never canceled on timeout. The code even acknowledges it on line 74-77: `The query goroutine is left running; if the terminal answers late, stray bytes may be consumed from stdin.`

"Stray bytes" is an understatement. It's all bytes, forever, and freezes the TUI. Changes requested.

rodboev:
Tested the fix below on Windows Terminal. It replaces the leaking goroutine with a synchronous `WaitForSingleObject` loop. No freeze, TUI responds to input normally:

```go
	if _, err := os.Stdout.WriteString("\x1b]11;?\x1b\\"); err != nil {
		return false, false
	}

	deadline := time.Now().Add(windowsBackgroundQueryTimeout)
	var buf []byte
	for time.Now().Before(deadline) {
		millis := uint32(time.Until(deadline).Milliseconds())
		if millis == 0 {
			break
		}
		event, _ := windows.WaitForSingleObject(stdin, millis)
		if event != windows.WAIT_OBJECT_0 {
			break
		}
		var tmp [256]byte
		n, err := os.Stdin.Read(tmp[:])
		if err != nil || n == 0 {
			break
		}
		buf = append(buf, tmp[:n]...)
		if bytes.ContainsAny(buf, "\a") || bytes.Contains(buf, []byte("\x1b\\")) {
			return parseOSCBackgroundIsDark(string(buf))
		}
	}
	return false, false
```

My reviews db got mangled, waiting on a few to populate so I can check the result of the PR.

rodboev:
This improves the readability of text that was showing up white-on-white before.

<table><tr><td>Before:
image</td><td>After:
image</td></tr></table>

The commands at the bottom are more washed out than before, but overall the TUI is more usable in light themed environments.

wesm:
looking

wesm:
rejected

PR #1198, opened 2026-09-17
fix(refine): initialize worktree submodules with the user's git config

### LINKED ISSUE #1195: Refine cannot initialize private HTTPS submodules because Git config isolation hides credential helpers
## Problem

`roborev refine` fails while creating its temporary worktree when the repository contains a private HTTPS submodule.

The submodule is accessible through ordinary Git because GitHub CLI is configured as the global credential helper. RoboRev runs the submodule update with isolated Git configuration, so that helper is unavailable. Since terminal prompts are also disabled, the clone fails before the fix agent starts.

```text
Error: create worktree: git submodule update: git submodule update --init: exit status 1
Cloning into .../roborev-worktree-.../api...
fatal: could not read Username for 'https://github.com': terminal prompts disabled
fatal: clone of 'https://github.com/lakehq/api-spec.git' into submodule path '.../api' failed
```

## Regression

This worked with RoboRev v0.56.0. PR #764 changed refine from the internal worktree helper to `gitworktree.Create` with `InitSubmodules: true`. Kit's Git runner sets an empty `GIT_CONFIG_GLOBAL`, sets `GIT_CONFIG_NOSYSTEM=1`, and disables terminal prompts. The combined behavior makes private HTTPS submodules impossible to initialize when authentication comes from a normal global credential helper.

Issue #888 reported the same authentication error, but PR #896 addressed modification and commit safety for submodules rather than authentication during worktree creation. The failure remains in v0.67.0.

## Reproduction

1. Configure GitHub CLI authentication with `gh auth setup-git`.
2. Confirm `GIT_TERMINAL_PROMPT=0 git ls-remote https://github.com/<org>/<private-repo>.git HEAD` succeeds.
3. Add that private repository as an HTTPS submodule in another repository.
4. Run `roborev refine`.
5. Observe that temporary-worktree creation fails during `git submodule update --init`.

## Expected behavior

Refine should initialize private submodules using the user's existing non-interactive Git credentials, without exposing credentials to agents or untrusted repository configuration. If RoboRev intentionally cannot forward credential helpers, it should provide a documented credential mechanism or an option to skip submodule initialization when the fix does not require submodule contents.

## Environment

- RoboRev v0.67.0
- macOS
- Private GitHub submodule over HTTPS
- Authentication configured through `gh auth setup-git`

rodboev:
`roborev refine` now initializes private HTTPS submodules in its temporary worktree with the Git credentials already configured by the operator. Worktree creation can reach code the fix agent needs before the agent starts.

Previously, refine hid the global and system Git configuration while initializing the worktree. That removed credential helpers and made `git submodule update --init` fail with terminal prompts disabled. Refine now reads persistent Git configuration while still stripping inherited `GIT_*` variables.

Git resolves its own helper and URL rules, so the change avoids copying a partial list of credential keys into the command. The daemon's background fix worktrees keep their isolated configuration.

Closes #1195


wesm:
rejected

